PR #9, opened 2021-12-18
3 setup ci pipeline docker based

gerardo-navarro:
## Checklist
- [ ] Work on docker build in ci cd workflow
- [ ] Reuse docker container for linter and test execution

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,18 +9,6 @@ on:
             - main
 
 jobs:
-    my_job:
-        runs-on: ubuntu-latest
-        container: node:14.16
-        steps:
-            - run: pwd
-            - run: ls -lisa
-            - run: ls -lisa /
-            - uses: actions/checkout@v2
-            - run: ls -lisa
-            - run: ls -lisa /
-
-
     mindmapper-build:
         runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+                      ghcr.io/b310-digital/mindmapper/ci:${{ github.sha }}-dev
+            #   outputs: type=docker,dest=/tmp/myimage.tar
+
+            # - name: Upload artifact
+            #   uses: actions/upload-artifact@v2
+            #   with:
+            #       name: myimage
+            #       path: /tmp/myimage.tar
+
+    mindmapper-backend-lint-2:
```
JannikStreek, line 47: why the name lint-2?

JannikStreek: ah ok so there are two lint jobs, one based on docker, one not? do you want to keep both or what's the idea here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Build and export to Docker
+              uses: docker/build-push-action@v2
+              with:
+                  context: .
+                  target: development
+                  push: true
+                  tags: |
+                      ghcr.io/b310-digital/mindmapper/ci:${{ github.sha }}-dev
```
JannikStreek, line 38: this image is now shown in the package section. Unfortunately, it is named only "mindmapper" and only by looking at the URL you can see, that it is actually the dev image. I am also wondering how this will work on branches or is this only triggered when deployed to main?

.github/workflows/ci.yml
gerardo-navarro: This branch is not ready yet.

The idea is to build the docker image during the ci co build (<= this works already). And then use this docker image to execute the other jobs, e.g. `lint` and 1test:e2e` <= I have to work on this a little bit further.

For the time being I am keeping the existing jobs. When I am done with the configuration of the ci cd pipeline. I will remove the others.

JannikStreek:
@gerardo-navarro can you have a look at this / finish it? I think it would still make change to improve the ci file.

PR #136, opened 2022-11-12
improve font size handling

### LINKED ISSUE #135: A better handling of the text size would be ingenious
Good mornng

**thank you very much for sharing your great work!**

I support small non-profit projects and I've tested TeamMapper for them.
I really like the tool, but because one can change the font size of the individual nodes individually by probably +/- 1 pixel, the diagrams look very quickly funny / choppy.

If one wants to make the diagram more professional and the font sizes more uniform, it becomes very tedious: When changing the font size, one does not see anywhere how big the font is exactly (e.g. "12px") and one has to estimate visually if it is approximately correct. And one must select each node individually to adjust the text size.

#### That's why I have these ideas to improve the handling of the font size:
- It would be great if one can see in the slider for the font size in a tooltip, which size one selects at the moment.
- It would be great if one could configure in the settings how big the step size is when changing the font size.
  I guess one can change the font size by +/- 1Px at the moment.
  Depending on how much detail my chart has, I would choose a larger increment.
- It would be even better if one could explicitly specify in the settings which font sizes are used, e.g. "8px, 12px, 15px, 17px".
  Then the slider would allow the user to choose only between these predefined sizes.

Thanks a lot, kind regards,
Thomas

JannikStreek:
closes #135 

Todos:

- [x] Hotkeys missing
- [x] Sync map settings with server to avoid inconsistent settings where one user uses 50px as max font size and another 70px
- [x] refactor and use default values instead of hardcoded ones
- [x] Refactor validator in new method

### DIFF
--- a/teammapper-backend/src/migrations/1668360651755-AddOptionsToMap.ts
+++ b/teammapper-backend/src/migrations/1668360651755-AddOptionsToMap.ts
@@ -1,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddOptionsToMap1668360651755 implements MigrationInterface {
+    name = 'AddOptionsToMap1668360651755'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "mmp_map" ADD "options" jsonb NOT NULL DEFAULT '{}'`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "mmp_map" DROP COLUMN "options"`);
+    }
+
+}
--- a/teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
+++ b/teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
@@ -220,7 +220,7 @@ export class MapSyncService {
 
   public async updateMapOptions (options?: CachedMapOptions): Promise<void> {
     const cachedMapEntry: CachedMapEntry = this.getAttachedMap()
-    this.socket.emit('updateMapOptions', { mapId: cachedMapEntry.cachedMap.uuid, options: options })
+    this.socket.emit('updateMapOptions', { mapId: cachedMapEntry.cachedMap.uuid, options: options })
   }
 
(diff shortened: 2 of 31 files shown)


teammapper-frontend/src/app/modules/application/pages/settings/settings.component.ts
```diff
@@ -39,4 +53,10 @@ export class SettingsComponent {
   public back () {
     this.location.back()
   }
+
+  private validateMapOptionsInput() {
+    if(this.mapOptions.fontIncrement > this.mapOptions.fontMaxSize || this.mapOptions.fontIncrement < 1) this.mapOptions.fontIncrement = 5
```
JannikStreek, line 58: these should come either from a json file or should be static vars

teammapper-backend/src/migrations/1668360651755-AddOptionsToMap.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddOptionsToMap1668360651755 implements MigrationInterface {
+    name = 'AddOptionsToMap1668360651755'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "mmp_map" ADD "options" jsonb NOT NULL DEFAULT '{}'`);
```
JannikStreek, line 7: without string

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
```diff
@@ -108,45 +108,54 @@ export class MapSyncService {
...
       return false
     }
 
-    const mapKey = this.createKey(newMap.uuid)
+    const mapKey = this.createKey(newServerMap.uuid)
 
     this.attachMap({
       key: mapKey,
-      cachedMap: newMap
+      cachedMap: { ...this.convertServerMapToMmp(newServerMap), ...{ options: newServerMap.options } }
```
JannikStreek, line 147: explain the difference between a mmp, server map and cached map somewhere

JannikStreek:
input fields are default empty, need initial data

PR #151, opened 2023-03-12
improve and reenable links

### LINKED ISSUE #69: Link creation with support for mobile safari
Links inside a foreign object element seem not to work on mobile safari. Therefore, the link creation strategy has to leverage native svg link elements instead of making use of foreign object. This requires bigger changes in the underlying map lib. The alternative would be to add an onclick element, but this leaves room for possible XSS exploits.

Using native svg links means, that not these node are not editable anymore.

JannikStreek:
closes #69 

- Add commented link button
- Add href property to map and backend
- show link below node
- improve node editing behavior on mobile devices
- legacy support
- remove link
-  readme


teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -1021,6 +1046,8 @@ export const PropertyMapping = {
     coordinates: ['coordinates'],
     imageSrc: ['image', 'src'],
     imageSize: ['image', 'size'],
+    linkName: ['link', 'name'],
```
JannikStreek, line 1049: remove

JannikStreek:
Missing: Remove link button option

PR #154, opened 2023-03-16
Use node lastModified  instead of map lastModified for cleanup

### LINKED ISSUE #109: use last node modifiedAt instead of map modifiedAt
Currently maps are deleted after 30 days. This is calculated based on the modified at attribute. Unfortunately, when nodes change, the map modifiedAt does not. Therefore, we should only delete maps, where the last node was modified 30 days ago.

Links:

- Typeorm documentation: https://typeorm.io
- Current query: https://github.com/b310-digital/teammapper/blob/6e7a67f32e5d4a0af418624838d7081981c3e297/teammapper-backend/src/map/services/maps.service.ts#L111
- Job: https://github.com/b310-digital/teammapper/blob/main/teammapper-backend/src/jobs/deleteMaps.job.ts
- Tests for the service methods: https://github.com/b310-digital/teammapper/blob/main/teammapper-backend/src/map/services/maps.service.spec.ts

nwittstruck:
Closes #109 

- [x] migration is missing

### DIFF
--- a/teammapper-backend/src/jobs/deleteMaps.job.ts
+++ b/teammapper-backend/src/jobs/deleteMaps.job.ts
@@ -14,8 +14,8 @@ async function bootstrap() {
   const mapsService = application.get(MapsService);
 
   logger.log('--- Deleting old maps ... ---');
-  const result: DeleteResult = await mapsService.deleteOutdatedMaps(configService.deleteAfterDays());
-  logger.log('Deleted rows: ' + result.affected);
+  const result: Number = await mapsService.deleteOutdatedMaps(configService.deleteAfterDays());
+  logger.log('Deleted rows: ' + result);
   logger.log('--- Finished deleting maps ---');
 
   await application.close();
--- a/teammapper-backend/src/jobs/deleteMaps.job.ts
+++ b/teammapper-backend/src/jobs/deleteMaps.job.ts
@@ -13,7 +13,7 @@ async function bootstrap() {
   const mapsService = application.get(MapsService);
 
   logger.log('--- Deleting old maps ... ---');
-  const result: Number = await mapsService.deleteOutdatedMaps(configService.deleteAfterDays());
+  const result: number = await mapsService.deleteOutdatedMaps(configService.deleteAfterDays());
   logger.log('Deleted rows: ' + result);
   logger.log('--- Finished deleting maps ---');
 
--- a/teammapper-backend/src/jobs/deleteMaps.job.ts
+++ b/teammapper-backend/src/jobs/deleteMaps.job.ts
@@ -3,7 +3,6 @@ import { MapsService } from '../map/services/maps.service';
 import AppModule from '../app.module';
 import { Logger } from '@nestjs/common';
 import configService from '../config.service';
-import { DeleteResult } from 'typeorm';
 
 async function bootstrap() {
   const application = await NestFactory.createApplicationContext(
--- a/teammapper-backend/src/map/entities/mmpNode.entity.ts
+++ b/teammapper-backend/src/map/entities/mmpNode.entity.ts
@@ -96,4 +96,7 @@ export class MmpNode {
   @Column({ nullable: false })
   @Generated('increment')
     orderNumber: number;
+
+  @Column({ type: 'timestamptz', nullable: true, default: () => 'now()' })
+    lastModified: Date;
 }
--- a/teammapper-backend/src/map/services/maps.service.spec.ts
+++ b/teammapper-backend/src/map/services/maps.service.spec.ts
@@ -38,6 +38,15 @@ describe('MapsController', () => {
     return truncateDatabase(mapsRepo.manager.connection)
   });
 
+  const createNode = async (map: MmpMap, lastModified: Date) => {
+    return nodesRepo.save({
+        nodeMapId: map.id,
+        coordinatesX: 3,
+        coordinatesY: 1,
+        lastModified: lastModified,
+      });
+  }
+
   describe('updateNode', () => {
     it('does update the lastModified value on update', async() => {
       const map: MmpMap = await mapsRepo.save({
@@ -45,12 +54,7 @@ describe('MapsController', () => {
       });
 
       const oldDate = new Date('2019-01-01')
-      const node: MmpNode = await nodesRepo.save({
-        nodeMapId: map.id,
-        coordinatesX: 3,
-        coordinatesY: 1,
-        lastModified: oldDate,
-      });
+      const node: MmpNode = await createNode(map, oldDate);
 
       const clientNode = mapMmpNodeToClient(node);
       clientNode.name = 'new';
@@ -79,12 +83,7 @@ describe('MapsController', () => {
         lastModified: new Date('2019-01-01'),
       });
 
-      const node: MmpNode = await nodesRepo.save({
-        nodeMapId: map.id,
-        coordinatesX: 3,
-        coordinatesY: 1,
-        lastModified: new Date('2019-01-01'),
-      });
+      const node: MmpNode = await createNode(map, new Date('2019-01-01'));
 
       await mapsService.deleteOutdatedMaps(30);
       expect(await mapsService.findMap(map.id)).toEqual(null);
@@ -97,12 +96,7 @@ describe('MapsController', () => {
         lastModified: new Date('2019-01-01'),
       });
 
-      const node: MmpNode = await nodesRepo.save({
-        nodeMapId: map.id,
-        coordinatesX: 3,
-        coordinatesY: 1,
-        lastModified: new Date(),
-      });
+      const node: MmpNode = await createNode(map, new Date());
 
       await mapsService.deleteOutdatedMaps(30);
       expect(await mapsService.findMap(map.id)).not.toBeNull();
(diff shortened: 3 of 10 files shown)


teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -107,20 +108,60 @@ export class MapsService {
...
     return copyDate;
   }
 
-  async deleteOutdatedMaps(afterDays: number = 30): Promise<DeleteResult> {
-    return this.mapsRepository
+  async deleteOutdatedMaps(afterDays: number = 30): Promise<Number> {
+    const today = new Date();
+
+    const deleteQuery = this.mapsRepository
       .createQueryBuilder()
```
JannikStreek, line 135: you can also name the relation here for followup queries, e.g. this.mapsRepository('mmpMaps')

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -107,20 +108,60 @@ export class MapsService {
...
+      .orWhere(new Brackets((qb) => {
+        // also delete empty maps, use th emaps lastmodified date for this:
+        qb.where("lastmodifiednode.lastUpdatedAt IS NULL")
+        .andWhere("(MmpMap.lastModified + (INTERVAL '1 day' * :afterDays)) < :today", { afterDays: afterDays, today: today})
+      }));
+      
+      const outdatedMapsIdsFlat = (await deleteQuery.getRawMany()).flatMap(id => id["MmpMap_id"]);
+
+      if (outdatedMapsIdsFlat.length > 0) {
+        return (await this.mapsRepository
```
JannikStreek, line 155: small change to get rid of native table naming:

```
 return (await this.mapsRepository
           .createQueryBuilder('mmpMap')
           .where("mmpMap.id IN (:...ids)", { ids: outdatedMapsIdsFlat })
           .delete()
           .execute())
           .affected;
```

teammapper-backend/src/map/entities/mmpNode.entity.ts
```diff
@@ -96,4 +96,7 @@ export class MmpNode {
   @Column({ nullable: false })
   @Generated('increment')
     orderNumber: number;
+
+  @Column({ type: 'timestamptz', nullable: true, default: () => 'now()' })
```
JannikStreek, line 100: just fyi: this change will add a default date for all existing nodes as well (affected by the migration). Should we only add it for NEW nodes as a default or is the other case fine? Its not a big problem I guess but old maps will than stay active for maybe 2 months after this migration in the worst case (because otherwise they would have been deleted tomorrow and with this change you reactivated them for another 30 days)

teammapper-backend/src/jobs/deleteMaps.job.ts
```diff
@@ -14,8 +13,8 @@ async function bootstrap() {
   const mapsService = application.get(MapsService);
 
   logger.log('--- Deleting old maps ... ---');
-  const result: DeleteResult = await mapsService.deleteOutdatedMaps(configService.deleteAfterDays());
-  logger.log('Deleted rows: ' + result.affected);
+  const result: Number = await mapsService.deleteOutdatedMaps(configService.deleteAfterDays());
```
JannikStreek, line 16: number is fine, you don't need the wrapper Number?

teammapper-backend/src/map/services/maps.service.spec.ts
```diff
@@ -33,9 +33,46 @@ describe('MapsController', () => {
...
+      expect(updatedNode.lastModified.getTime()).toBeGreaterThan(timeBeforeUpdate.getTime());
+    });
+  });
+
   describe('deleteOutdatedMaps', () => {
-    it('deletes the old map and node', async () => {
+    it('does not delete a new map', async () => {
+
+      const map: MmpMap = await mapsRepo.save({
+        lastModified: new Date(),
```
JannikStreek, line 67: maybe also add a test that creates a new map via the service? otherwise it could be forgotten there to set the lastModified field?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -107,20 +108,60 @@ export class MapsService {
     return await this.mapsRepository.findOne({ where: { id: mapId }} );
   }
 
-  getDeletedAt(lastModified: Date, afterDays: number): Date {
+  async getDeletedAt(map: MmpMap, afterDays: number): Promise<Date> {
+    // get newest node of this map:
+    const newestNodeQuery = this.nodesRepository
+                                .createQueryBuilder()
+                                .select("max(MmpNode.lastModified) AS lastModified")
```
JannikStreek, line 115: same here, if you add a string parameter to the query builder, I think you can use it as an alias.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -107,20 +108,60 @@ export class MapsService {
...
+                                .where( { nodeMapId: map.id } )
+    const newestNode = newestNodeQuery.getRawOne()
```
JannikStreek, line 117: why not ` .getOne()` ? that would also enable to use a proper object accessor instead of hash syntax

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -107,20 +108,60 @@ export class MapsService {
...
+      .select("MmpMap.id")
+      .leftJoin(qb => 
+        // subquery to get the newest node and its lastModified date of this map:
+        qb
+        .select(["node.nodeMapId AS nodeMapId", "max(node.lastModified) AS lastUpdatedAt"])
+        .from("MmpNode", "node")
+        .groupBy("node.nodeMapId"), "lastmodifiednode", "lastmodifiednode.nodeMapid = MmpMap.id")
+      .where(
+        // delete all maps that have nodes that were last updated after afterDays
+        "(lastmodifiednode.lastUpdatedAt + (INTERVAL '1 day' * :afterDays)) < :today", { afterDays: afterDays, today: today})
```
JannikStreek, line 145: `{ afterDays: afterDays, today: today}` can be written as `{ afterDays, today }`

teammapper-backend/src/map/services/maps.service.ts
JannikStreek: https://orkhan.gitbook.io/typeorm/docs/select-query-builder#how-to-create-and-use-a-querybuilder

JannikStreek:
```

const user = await dataSource.manager
    .createQueryBuilder(User, "user")
    .where("user.id = :id", { id: 1 })
    .getOne()
```

or 

```

const user = await dataSource
    .getRepository(User)
    .createQueryBuilder("user")
    .where("user.id = :id", { id: 1 })
    .getOne()
```

teammapper-backend/src/map/services/maps.service.ts
nwittstruck: Unfortunately this doesn't work, it complains that the FROM clause is missing. 
```
 return (await this.mapsRepository
           .createQueryBuilder('mmpMap')
           .from(MmpMap, 'mmpMap')
           .where("mmpMap.id IN (:...ids)", { ids: outdatedMapsIdsFlat })
           .delete()
           .execute())
           .affected;
```

Adding it doesn't help either. This is a bit weird, isn't it?

teammapper-backend/src/map/entities/mmpNode.entity.ts
nwittstruck: I think this is fine, since we don't know when the previous nodes were created. They could be a day old or 29 days.

teammapper-backend/src/map/services/maps.service.ts
nwittstruck: unfortunately getOne doesn't work with max().

JannikStreek: if both are valid options for you, I would prefer the one with `getOne` as you can use normal object selectors after that - but if you think max is better, its also fine.

teammapper-backend/test/db.ts
```diff
@@ -1,20 +1,78 @@
...
+  await connection.query(`DROP DATABASE IF EXISTS "${databaseName}"`);
+  await connection.query(`CREATE DATABASE "${databaseName}"`);
+  await connection.destroy();
+
+  return databaseName;
+}
+
+// this is the configuration for the main test database. this database is not used for actual tests.
+// it is the entrypoint for each worker to be able to create their own worker database
+const mainTestDataSource = async () => {
```
JannikStreek, line 19: i understand that there seem to be some differences between these two test configs (e.g. sync). Still I would propose not copy paste it here, but merge it - e.g.:

```
const connection = new DataSource({ ... workerDataSourceConfig, synchronize: false })
``` 

> we don't need to synchronise the tables for the main test database

Also i am still not 100% sure why now workers (`workerDataSourceConfig`) use synchronize true, but the main config sets this to false. Also you would need then to run migrations if this is set to false, no?

teammapper-backend/src/map/services/maps.service.spec.ts
```diff
@@ -7,19 +7,21 @@ import { MmpNode } from '../entities/mmpNode.entity';
...
   let mapsRepo: Repository<MmpMap>;
+  let moduleFixture: TestingModule;
 
   beforeAll(async () => {
-    const moduleFixture: TestingModule = await Test.createTestingModule({
+    moduleFixture = await Test.createTestingModule({
       imports: [
         ConfigModule,
-        TypeOrmModule.forRoot(createTestConfiguration()),
+        TypeOrmModule.forRoot(await createTestConfiguration(process.env.JEST_WORKER_ID)),
```
JannikStreek, line 24: where is the env variable coming from, do we have to set this in the environment or is the framework managing the id?

teammapper-backend/test/db.ts
nwittstruck: Yes, I'll extract that to a function. The reason we don't run migrations in the main test database is, that no tests are running against this database. This database is just required to be able to initially connect to postgres, to then create additional databases. You cannot connect to postgres without a database.

PR #160, opened 2023-03-22
WIP: add cypress

### LINKED ISSUE #153: introduce e2e test with cypress
Unfortunately the frontend does not include any tests yet. A start would be e2e.

Example tests:

- [ ] Visit the map, add a node with a name
- [ ] Visit the map, invoke share button
- [ ] Visit the map, add a node, increase font size
- [ ] Visit the map, add a node, decrease font size
- [ ] Visit the map, add a node and add a color
- [ ] Visit the map, add two nodes, cut one node an append it to the other
- [ ] Visit the map, import an existing map

JannikStreek:
closes #153 

- [ ] Integrate cypress as image in docker compose setup
- [ ] optional: provide a separate docker compose file which leverages different databases to not dump data in dev db
- [ ] Add a couple of example tests, for ideas see issue
- [ ] idea: there should be a e2e folder directly in the root directory, where cypress data is stored and leverages

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       DELETE_AFTER_DAYS: ${DELETE_AFTER_DAYS:-30}
     ports:
       - "${APP_FRONTEND_PORT:-4200}:4200"
-      - "${APP_BACKTEND_PORT:-3000}:3000"
+      - "${APP_BACKEND_PORT:-3000}:3000"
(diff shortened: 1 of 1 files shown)


JannikStreek:
FYI: I added chromium to the depedencies on the main branch, it should be possible now to execute tests in headless mode with chromium directly in the app container.

PR #162, opened 2023-03-22
WIP: inline fonts in svg

### LINKED ISSUE #161: Font inside SVGs not displayed
Due to relative paths / missing fonts, the text inside the exported SVG is empty.

JannikStreek:


JannikStreek:
closing for now - the problem is not the missing inlined font, the problem is the foreign object. Therefore, this works but does not add value currently and only bloats the svg.

PR #285, opened 2024-06-15
#284 Add a copy mind map button

### LINKED ISSUE #284: Add a copy mind map button
**Proposed feature**
Currently, maps can be shared to enable multi user sessions. However, currently it is not possible to copy a mind map with a similar mechanism. The idea would be: A button in the share modal with which you can create a new map with the same content as present in the current map. This enables a fast copy instead of exporting the Map, creating a new map and importing content. 

- [ ] Think about where to put the button in the share modal
- [ ] A click creates a new mind map with the same content (in a new tab?)

**Additional context**
[Ecalidraw](https://excalidraw.com) has a similar feature as a reference. Click on "save as" and then on as a link.

Bildschirmfoto 2024-06-04 um 15 34 37

sorenjohanson:
closes #284


JannikStreek:
thanks, already looking good 👍 

UX:
- Maybe we can add a small info note popup that says "Successfully cloned" ? Otherwise it's the user might think nothing happened - as the map looks the same. Only thing I could found is the Dialog in material ui.

PR:
- Please Add a "closes #ISSUE_NUMBER_HERE" in the description

Test:
- Maybe add one controller test, only for this method? we can add other controller tests in a separate PR.

teammapper-backend/src/map/controllers/maps.controller.ts
```diff
@@ -49,4 +49,27 @@ export default class MapsController {
       modificationSecret: newMap.modificationSecret,
     }
   }
+
+  @Post(':id/share')
```
JannikStreek, line 53: duplicate instead of share?

teammapper-backend/src/map/controllers/maps.controller.ts
```diff
@@ -49,4 +49,27 @@ export default class MapsController {
...
+  async duplicate(
+    @Param('id') mapId: string,
+  ): Promise<IMmpClientPrivateMap> {
+    const oldMap = await this.mapsService.findMap(mapId).catch((e: Error) => {
+      if (e.name === 'MalformedUUIDError') throw new NotFoundException()
+    })
+
+    if (!oldMap) return Promise.reject()
```
JannikStreek, line 61: not found exception might be better suited here

sorenjohanson:
> UX:
> 
> * Maybe we can add a small info note popup that says "Successfully cloned" ? Otherwise the user might think nothing happened - as the map looks the same. Only thing I could found is the Dialog in material ui.

Do we already have something similar for other use cases? Could integrate ngx-toastr otherwise, but I don't know if that's maybe a bit overkill.

teammapper-backend/src/map/controllers/maps.controller.ts
sorenjohanson: 👍

teammapper-backend/src/map/controllers/maps.controller.ts
sorenjohanson: 👍

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -54,45 +54,56 @@ export class MapsService {
...
 
-    const newNode = this.nodesRepository.create({
-      ...mapClientNodeToMmpNode(clientNode, mapId),
-      nodeMapId: mapId,
-    })
+    let newNode;
+    // Type guard: IMmpClientNode
+    if ('parent' in clientNode) {
+      newNode = this.nodesRepository.create({
+        ...mapClientNodeToMmpNode(clientNode, mapId),
```
JannikStreek, line 72: I would propose to call this method (the mapping method) before calling "addNode" so we could use just one type (MmpNode) instead of the union type. it would be much more straightforward than making all these case switches.

sorenjohanson: Good point, would simplify the validatesNodeParentForNode() method as well, but we'd have to move the case switches to addNodes()

JannikStreek:
> > UX:
> > 
> > * Maybe we can add a small info note popup that says "Successfully cloned" ? Otherwise the user might think nothing happened - as the map looks the same. Only thing I could found is the Dialog in material ui.
> 
> Do we already have something similar for other use cases? Could integrate ngx-toastr otherwise, but I don't know if that's maybe a bit overkill.

only material dialogs which are also quite heavy for just displaying a simply message. otherwise we could also just code/style or own, not sure if a library is suited just for small alerts.

teammapper-backend/src/map/services/maps.service.ts
JannikStreek: well I would do the mapping then even before addNodes - so we also just have one type here?

sorenjohanson: The problem with that is that addNodes is called both from backend (with MmpNode) and extensively from the frontend (through websocket), so we'd end up doing the mapping in a bunch of different places instead.

JannikStreek: > The problem with that is that addNodes is called both from backend (with MmpNode) and extensively from the frontend (through websocket), so we'd end up doing the mapping in a bunch of different places instead.

then I would suggest to add a little helper method like `addNodesFromClient` which itself maps the data and then calls `addNodes`?

sorenjohanson: 👍

teammapper-backend/src/map/utils/clientServerMapping.ts
```diff
@@ -73,7 +73,7 @@ const mapClientNodeToMmpNode = (
   locked: clientNode.locked,
   detached: clientNode.detached,
   name: clientNode.name,
-  nodeParentId: clientNode.parent ? clientNode.parent : null,
+  nodeParentId: clientNode.parent ? clientNode.parent : undefined,
```
JannikStreek, line 76: `clientNode.parent` is enough now

teammapper-backend/src/map/utils/clientServerMapping.ts
```diff
@@ -95,7 +95,7 @@ const mapClientBasicNodeToMmpRootNode = (
   imageSrc: clientRootNodeBasics.image?.src,
   imageSize: clientRootNodeBasics.image?.size,
   name: clientRootNodeBasics.name || DEFAULT_NAME,
-  nodeParentId: null,
+  nodeParentId: undefined,
```
JannikStreek, line 98: I think we can remove this line, as a missing key should be undefined in the entity

teammapper-backend/src/map/controllers/maps.controller.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+                fontMinSize: 1,
+                fontIncrement: 1
+            },
+        };
+        const result: IMmpClientPrivateMap = {
+            map: exportedMap,
+            adminId: 'new-admin-id',
+            modificationSecret: 'new-modification-secret'
+        };
+        const oldNodes: [] = [];
```
JannikStreek, line 79: type is not needed here

teammapper-backend/src/map/controllers/maps.controller.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        expect(mapsService.findMap).toHaveBeenCalledWith(oldMap.id);
+        expect(mapsService.createEmptyMap).toHaveBeenCalled();
+        expect(mapsService.findNodes).toHaveBeenCalledWith(oldMap.id);
+        expect(mapsService.addNodes).toHaveBeenCalledWith(mapId, oldNodes);
+        expect(mapsService.exportMapToClient).toHaveBeenCalledWith(mapId);
+        expect(response).toEqual(result);
+
+        expect(newMap.name).toEqual(oldMap.name);
+        expect(newMap.lastModified).toEqual(oldMap.lastModified);
+        expect(newMap.nodes).toEqual(oldMap.nodes);
```
JannikStreek, line 98: have you verified that this is actually an array with nodes? be careful calling this without having specified that typeorm will load external relationships. as a default it won't and might just compare empty arrays here

teammapper-backend/src/map/controllers/maps.controller.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        jest.spyOn(mapsService, 'createEmptyMap').mockResolvedValueOnce(newMap);
+        jest.spyOn(mapsService, 'findNodes').mockResolvedValueOnce(oldNodes);
+        jest.spyOn(mapsService, 'addNodes').mockResolvedValueOnce([]);
+        jest.spyOn(mapsService, 'exportMapToClient').mockResolvedValueOnce(exportedMap);
+
+        const response = await mapsController.duplicate(oldMap.id);
+
+        expect(mapsService.findMap).toHaveBeenCalledWith(oldMap.id);
+        expect(mapsService.createEmptyMap).toHaveBeenCalled();
+        expect(mapsService.findNodes).toHaveBeenCalledWith(oldMap.id);
```
JannikStreek, line 91: thats a lot of expectations and this is already a long test. I would prefer just testing the end result which leads to the conclusion that all methods calls before worked. Otherwise I would prefer it to split up some exactions into a separate test. You can use a before block to reuse created data.

teammapper-backend/src/map/utils/clientServerMapping.ts
sorenjohanson: Yeah, removed.

teammapper-backend/src/map/utils/clientServerMapping.ts
sorenjohanson: Changed

teammapper-backend/src/map/controllers/maps.controller.spec.ts
sorenjohanson: Removed.

teammapper-backend/src/map/utils/clientServerMapping.ts
```diff
@@ -73,7 +73,7 @@ const mapClientNodeToMmpNode = (
   locked: clientNode.locked,
   detached: clientNode.detached,
   name: clientNode.name,
-  nodeParentId: clientNode.parent ? clientNode.parent : null,
+  nodeParentId: clientNode.parent || undefined,
```
JannikStreek, line 76: a comment would be nice here, as this looks rather strange on first glance :-D

PR #294, opened 2024-06-19
Add toasts where sensible

### LINKED ISSUE #288: Add toasts where sensible
**Proposed feature**
In #284, we've added the ngx-toastr package to display a toast when successfully duplicating a mindmap. We should check where else toasts are sensible and add them.

sorenjohanson:
closes #288

### DIFF
--- a/teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
+++ b/teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
@@ -1,6 +1,7 @@
 import { Injectable, OnDestroy } from '@angular/core';
 import { Observable, Subscription } from 'rxjs';
 import { SettingsService } from '../settings/settings.service';
+import { ToastrService } from 'ngx-toastr';
 import { UtilsService } from '../utils/utils.service';
 import { jsPDF } from 'jspdf';
 import { first } from 'rxjs/operators';
@@ -31,7 +32,7 @@ export class MmpService implements OnDestroy {
   private additionalOptions: CachedMapOptions;
   private settingsSubscription: Subscription;
 
-  constructor(public settingsService: SettingsService) {
+  constructor(public settingsService: SettingsService, public utilsService: UtilsService, public toastrService: ToastrService) {
     this.additionalOptions = null;
     this.branchColors = COLORS;
 
@@ -314,8 +315,21 @@ export class MmpService implements OnDestroy {
    * Copy a node with his children in the mmp clipboard.
    * If id is not specified, copy the selected node.
    */
-  public copyNode(nodeId?: string) {
-    this.currentMap.instance.copyNode(nodeId);
+  public async copyNode(nodeId?: string) {
+    try {
+      this.currentMap.instance.copyNode(nodeId);
+
+      const successMessage = await this.utilsService.translate('TOASTS.NODE_COPIED');
+      this.toastrService.success(successMessage);
+    } catch (e) {
+      if (e.message == 'The root node can not be copied') {
+        const rootNodeFailureMessage = await this.utilsService.translate('TOASTS.ERRORS.ROOT_NODE_COPIED');
+        this.toastrService.error(rootNodeFailureMessage);
+      } else {
+        const genericErrorMessage = await this.utilsService.translate('TOASTS.ERRORS.GENERIC_NODE_COPY_ERROR');
+        this.toastrService.error(genericErrorMessage);
+      }
+    }
   }
 
   /**
--- a/teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
+++ b/teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
@@ -32,7 +32,11 @@ export class MmpService implements OnDestroy {
   private additionalOptions: CachedMapOptions;
   private settingsSubscription: Subscription;
 
-  constructor(public settingsService: SettingsService, public utilsService: UtilsService, public toastrService: ToastrService) {
+  constructor(
+    public settingsService: SettingsService,
+    public utilsService: UtilsService,
+    public toastrService: ToastrService
+  ) {
     this.additionalOptions = null;
     this.branchColors = COLORS;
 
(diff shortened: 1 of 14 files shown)


teammapper-frontend/src/assets/i18n/de.json
```diff
@@ -162,7 +162,14 @@
     "JSON_FILE_SIZE_TOO_LARGE": "Die Datei ist größer als 1 MB und kann möglicherweise später nicht importiert werden. Bitte Datei verkleinern (z.B. Bilder entfernen)."
   },
   "TOASTS": {
-    "SUCCESSFULLY_DUPLICATED": "Mindmap erfolgreich dupliziert! Leite weiter..."
+    "SUCCESSFULLY_DUPLICATED": "Mindmap erfolgreich dupliziert!",
+    "URL_COPIED": "URL erfolgreich kopiert!",
+    "NODE_COPIED": "Knoten erfolgreich kopiert!",
+    "ERRORS": {
+      "ROOT_NODE_COPIED": "Es ist nicht möglich, den Hauptknoten zu kopieren!",
+      "GENERIC_NODE_COPY_ERROR": "Beim kopieren des Knoten ist ein Fehler aufgetreten."
```
JannikStreek, line 170: > Beim Kopieren des Knotens

teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
```diff
@@ -314,8 +319,27 @@ export class MmpService implements OnDestroy {
...
+  public async copyNode(nodeId?: string) {
+    try {
+      this.currentMap.instance.copyNode(nodeId);
+
+      const successMessage = await this.utilsService.translate(
+        'TOASTS.NODE_COPIED'
+      );
+      this.toastrService.success(successMessage);
+    } catch (e) {
+      if (e.message == 'The root node can not be copied') {
```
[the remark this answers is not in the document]
JannikStreek: do we have to check for the message or isn't there a specific class / type field available?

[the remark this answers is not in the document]
JannikStreek: if not, maybe its a good idea to at least refactor this message as a constant variable. Better in the mmp lib, otherwise here at the top of the file

teammapper-frontend/src/app/guards/toast.guard.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(private router: Router, private toastrService: ToastrService) {}
+
+  canActivate(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ): Promise<boolean> | boolean {
+    const showToast = next.queryParamMap.get('showToast');
+    const toastMessage = next.queryParamMap.get('toastMessage');
+
+    if (showToast === 'true' && toastMessage) {
```
JannikStreek, line 23: instead of true/false as strings, maybe better to use 1/0?

JannikStreek: wouldnt be just the availability of a toastMessage enough? Or what's the case where I am passing a toast message without the showToast flag?

teammapper-frontend/src/app/guards/toast.guard.ts
```diff
@@ -0,0 +1,36 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class ToastGuard implements CanActivate {
+  constructor(private router: Router, private toastrService: ToastrService) {}
+
+  canActivate(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ): Promise<boolean> | boolean {
```
JannikStreek, line 19: only use one type, e.g. `Promise<boolean>`. I don't think you need to change anything else, otherwise you can always a return a `Promise.resolve(true)`.

teammapper-frontend/src/app/modules/application/components/dialog-about/dialog-about.component.ts
```diff
@@ -35,7 +35,18 @@ export class DialogAboutComponent {
     if (confirm(this.translateService.instant('MODALS.INFO.CONFIRM_DELETE'))) {
       await this.mapSyncService.deleteMap(await this.mapAdminId);
       await this.storageService.remove(this.map.uuid);
-      window.location.reload();
+
+      const url = new URL(window.location.href);
```
JannikStreek, line 39: can't we use the router.state here as well? I think its a more unified way of accessing the url

teammapper-frontend/src/app/modules/application/components/dialog-share/dialog-share.component.ts
```diff
@@ -72,20 +78,28 @@ export class DialogShareComponent implements OnInit {
...
+      const fragment = newMap.modificationSecret;
+
+      const url = new URL(baseUrl, window.location.origin);
+
+      url.searchParams.set('showToast', 'true');
+      url.searchParams.set('toastMessage', successMessage);
+
+      url.hash = fragment;
+
+      window.location.replace(url.toString());
```
JannikStreek, line 102: same here, maybe use angular router instead of doing this manually?

JannikStreek:
This is really helpful, I really like these messages where it was unclear before what really happened 😂 👍

teammapper-frontend/src/assets/i18n/de.json
JannikStreek: "Kopieren" needs still a capitalized K ;-)

teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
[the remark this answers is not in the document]
sorenjohanson: Don't think this has a specific type, but strict equality didn't work for some reason (even though e.message should be typeof string, but maybe it's JS doing JS things), was my first preference as well.

Question is if we've got any other errors this could potentially throw - right now I'm just seeing a specific error message when trying to copy the root node, otherwise it doesn't actually error out at all. If we had multiple errors I could see the value in a constant, but not sure if it's worth it for a one-off.

teammapper-frontend/src/app/guards/toast.guard.ts
sorenjohanson: Yeah good point about the toastMessage actually :D

teammapper-frontend/src/app/modules/application/components/dialog-about/dialog-about.component.ts
sorenjohanson: Yeah that was my preferred way of doing it since `router.navigate` is a much cleaner way of doing things and allows us to separate URL, query params and fragment, but wasn't sure if it was worth it for a one off (esp. considering the component doesn't have the router yet)

teammapper-frontend/src/assets/i18n/de.json
sorenjohanson: Added :D

sorenjohanson:
Latest commit fixes #297

teammapper-frontend/src/app/modules/application/components/dialog-share/dialog-share.component.ts
sorenjohanson: Added in now, the trick is to close the dialog manually before this.router.navigate()

teammapper-frontend/src/app/modules/application/components/dialog-about/dialog-about.component.ts
sorenjohanson: Added in this.router.navigate()

JannikStreek:
fyi there is still a open file conflict I think

sorenjohanson:
> fyi there is still a open file conflict I think

Merged in latest main now, resolved it hopefully.

sorenjohanson:
Looks like adding the DialogService causes Chrome Headless to crash during tests for some reason - works in practice though.

PR #296, opened 2024-06-20
Add lastAccessed field to map table

### LINKED ISSUE #286: Add lastAccessed field to map table
**Proposed feature**
We would like to standardize or deletion policies of old data across our open source apps. To make it as easy as possible, we have decided to leverage the last access of a created entity. Currently teammapper does not have such a field on the map entity and we use the lastModified field.

- [ ] Add a lastAccessed timestamp field to map entity
- [ ] The field should get updated when `findOne` is called in the controller => Please check that this method is actually called when accessing a map, I think it should be called.
- [ ] Adapt the deletion algorithm in the map service to use this field instead of lastModifiedAt and as a backup the old field (as many legacy maps don't have any data inside the new field).

sorenjohanson:
closes #286


teammapper-backend/src/map/entities/mmpMap.entity.ts
```diff
@@ -16,6 +16,9 @@ export class MmpMap {
   @Column({ type: 'timestamptz', nullable: true, default: () => 'now()' })
   lastModified: Date
 
+  @Column({ type: 'timestamptz', nullable: true })
```
JannikStreek, line 19: maybe also add the default here like in lastModified

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -37,6 +37,10 @@ export class MapsService {
     })
   }
 
+  async updateLastAccessed(uuid: string, lastAccessed = new Date()) {
+    await this.mapsRepository.update(uuid, { lastAccessed })
```
JannikStreek, line 41: `return this.mapsRepository.update(uuid, { lastAccessed })` and add the return type, e.g. Promise<void> (not sure what update returns)

teammapper-backend/src/map/entities/mmpMap.entity.ts
JannikStreek: on second thought, maybe better not to use the default, as all existing maps will then also get a lastAccessed for the actual time? Could you verify this?

sorenjohanson: Yeah my thought was just because a map is created it isn't immediately accessed (but it _is_ immediately modified), so I didn't add a default.

JannikStreek:
Please don't forget to generate the migration file for this, see https://github.com/b310-digital/teammapper/blob/main/teammapper-backend/README.md.

teammapper-backend/src/map/services/maps.service.ts
sorenjohanson: I think update returns something like UpdateResult which includes the affected row. Question is if we actually need that return value? When using deleteMap() (I think) we just call `this.mapsRepository.delete` and return void instead.

JannikStreek:
please also add the use of this field to the delete handler in the service (`deleteOutdatedMaps`), so that we use the lastAccessed field on map as a default and as a backup the lastModified field (existing logic). We can also do this in a new PR?

sorenjohanson:
> please also add the use of this field to the delete handler in the service (`deleteOutdatedMaps`), so that we use the lastAccessed field on map as a default and as a backup the lastModified field (existing logic). We can also do this in a new PR?

I'm working on that right now, it's just that I need to restructure the queries a bit to work with the new field, so it'll take some time.

teammapper-backend/src/map/services/maps.service.spec.ts
```diff
@@ -118,6 +118,15 @@ describe('MapsController', () => {
   })
 
   describe('deleteOutdatedMaps', () => {
+    it('deletes a map based off of lastAccessed', async () => {
```
JannikStreek, line 121: please also add test cases for these scenarios:

1) last access is set, but it should not be deleted due to more recent date
2) last access is not set at all, therefore old code should be used => lastModified
2a) lastModified too old => delete
2b) lastModified new => do not delete 
3) lastModified is too old but last access is rather new => should not be deleted
4) theoretical case, still might be sensible to test: last modified is new, lastAccessed too old => do not delete

JannikStreek: this is important because we have a running system with lots of data, would be rather unfortunate to loose all data ;)

teammapper-backend/src/map/controllers/maps.controller.ts
```diff
@@ -26,6 +26,8 @@ export default class MapsController {
     })
     if (!map) throw new NotFoundException()
 
+    await this.mapsService.updateLastAccessed(mapId)
```
JannikStreek, line 29: maybe refetch the map here, as this return will not include the lastAccess field on the map. Please also check that it is returned, as I think we will need it later on. E.g. the settings page for your own maps might use it to show the correct deletion date.

sorenjohanson: I did it the other way around: Updated lastAccessed first, then exportMapToClient is guaranteed to return a freshly updated lastAccessed field

teammapper-backend/src/map/services/maps.service.spec.ts
sorenjohanson: Good catch with point 3 and 4, both of these tests are failing right now :D Will fix tomorrow

teammapper-backend/src/jobs/seedMapData.job.ts
```diff
@@ -40,6 +40,7 @@ const createMap = (nodes: IMmpClientNode[]): IMmpClientMap => {
   return {
     uuid: crypto.randomUUID(),
     lastModified: new Date(),
+    lastAccessed: new Date(), // Do we really want to set this here? Without it there's an error
```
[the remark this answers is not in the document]
sorenjohanson: If I remember correctly it throws an error since it expects all attributes, and lastAccessed isn't optional (in TypeScript's eyes anyway), but if it's only test data we might as well leave it in there.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -233,15 +239,13 @@ export class MapsService {
...
         "(lastmodifiednode.lastUpdatedAt + (INTERVAL '1 day' * :afterDays)) < :today",
         { afterDays, today }
       )
       .orWhere(
         new Brackets((qb) => {
-          // also delete empty maps, use th emaps lastmodified date for this:
-          qb.where('lastmodifiednode.lastUpdatedAt IS NULL').andWhere(
-            "(map.lastModified + (INTERVAL '1 day' * :afterDays)) < :today",
+          qb.where("lastmodifiednode.lastUpdatedAt IS NULL").andWhere(
+            "(GREATEST(map.lastAccessed, map.lastModified) + (INTERVAL '1 day' * :afterDays)) < :today",
```
JannikStreek, line 248: is the "and" here correct? Usually, there shouldn't be any map without a node, therefore the first clause will be never null. So we might use it to collect broken maps but then it should be an or? Can you test this thought in a test case? I think most of the test do not include any nodes, so we are not really testing what is happening on real systems.

sorenjohanson: I've added two new test cases now:

- It should delete a map where both lastAccessed and lastUpdatedAt are outdated
- It should not delete a map where lastAccessed is outdated but lastUpdatedAt on some nodes are recent

Both work just fine. But I agree in theory that it feels like it's at the wrong spot and the first clause should never be true.

JannikStreek:
ok there are actually >300 maps on our server, that do not have any nodes. Interestingly, all these maps are not old but where created in September - so there seems to be a bug not creating the root node or deleting the root node.

But this is a bug for a different issue. However, we should still include this case, that a map does not contain any nodes. => `nodeLastUpdatedAt = null AND GREATER(map.lastModified, map.lastAccessed) ...`

The following test cases are the most relevant:
- Only lastModified is set on a map (either too old or recent enough) with one node that is was either modified before and after the lastModified date.
- lastModified AND lastAccess are set (one being older than the other) + with one node existing, being modified at a later point and at a newer point in time
- A map that has no nodes at all for some reason => only look at lastModified as a reference. I also thought about deleting those directly, but for debugging purposes it makes sense to keep them for some time.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -233,19 +239,10 @@ export class MapsService {
...
-      .orWhere(
-        new Brackets((qb) => {
-          // also delete empty maps, use th emaps lastmodified date for this:
-          qb.where('lastmodifiednode.lastUpdatedAt IS NULL').andWhere(
-            "(map.lastModified + (INTERVAL '1 day' * :afterDays)) < :today",
-            { afterDays, today }
-          )
-        })
-      )
+    // One case left: If lastModified is set but lastAccessed isn't and lastUpdatedAt is null, map should be deleted
```
JannikStreek, line 245: so this is a todo?

sorenjohanson: I accidentally left that in there, that's covered now.

teammapper-backend/src/map/services/maps.service.spec.ts
```diff
@@ -119,8 +119,64 @@ describe('MapsController', () => {
...
```
JannikStreek, line 122: typo: deletes a map based on `lastAccessed`. 

In addition: both values are way older than 30 days. I suggest to either lock the current date and set it to +1 day of lastAccessed or to make the dates relative, e.g. today - x days during creation.

JannikStreek: the date suggestion is true for basically all test, or am I missing something?

teammapper-backend/src/map/services/maps.service.spec.ts
```diff
@@ -152,6 +208,33 @@ describe('MapsController', () => {
       expect(await nodesRepo.findOne({ where: { id: node.id } })).not.toBeNull()
     })
 
+    it('deletes a map which has outdated nodes and outdated lastAccess', async () => {
```
JannikStreek, line 211: also add a test (if not given, can't find it right now): Outdated modifedAt, recent nodes => do not delete. This is how its working right now.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -233,19 +239,9 @@ export class MapsService {
         'lastmodifiednode.nodeMapid = map.id'
       )
       .where(
-        // delete all maps that have nodes that were last updated after afterDays
-        "(lastmodifiednode.lastUpdatedAt + (INTERVAL '1 day' * :afterDays)) < :today",
+        "(GREATEST(map.lastAccessed, map.lastModified, lastmodifiednode.lastUpdatedAt) + (INTERVAL '1 day' * :afterDays)) < :today",
```
JannikStreek, line 242: thats a cool discovery that this takes more than 2 arguments, makes it way easier 👍

teammapper-backend/src/map/services/maps.service.spec.ts
JannikStreek: e.g. https://stackoverflow.com/a/63377110

sorenjohanson: Not sure it'd make sense for this test? If we lock the date to +1 of lastAccessed, the map wouldn't be deleted, since it's not older than 30 days. The other tests circumvent this by setting `new Date()` explicitly.

We could set each date explicitly using jest to make it more clear, but I don't think it would directly affect the tests.

teammapper-backend/src/map/services/maps.service.spec.ts
sorenjohanson: This is covered by `it('does not delete a map that contains a recent node')`, where lastAccessed is set to 2019 but the node is set to today using `new Date()`.

teammapper-backend/src/map/services/maps.service.spec.ts
JannikStreek: I think it would help a lot to lock the date to a specific date (see above link with jest) so it's easier to understand what the tests are actually testing. Otherwise we might accidentally test something completely different than we thought. 

And I am referring to all tests, not necessarily to this specific one. These are all tests for which the correct date is crucial.

sorenjohanson:
Dates are now set explicitly in every test, in addition to the following format changes:

An "old" date is _always_ `2021-01-01`.
A "recent" date is _always_ `2024-09-01`.

If we're testing "recent" dates, system time will always be set exactly to the same date.
On the other hand, if we're testing "old" dates, system time will always be set to "old" date + 30 days, which will always result in system time being set to `2021-01-31`.

JannikStreek:
looks good, I also tested everything locally with cross running every minute and 1 day as delete threshold. Could not find any problems.

PR #300, opened 2024-06-24
Download bug when images exceed browser canvas limitations

### LINKED ISSUE #274: iPad image download bug for huge maps
**Describe the bug**
When exporting a big map (meaning not many nodes but just nodes with huge distances), image download on the iPad adds a `txt` ending which makes image usage difficult.

**To Reproduce**
1. Create a big mind map
2. Download as image on the iPad
3. See txt ending of file

**Expected behavior**
No txt ending of the file

sorenjohanson:
closes #274

### DIFF
--- a/teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
+++ b/teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
@@ -471,12 +471,20 @@ export class MmpService implements OnDestroy {
       case 'png': {
         const image = await this.exportAsImage(format);
 
-        UtilsService.downloadFile(
-          `${name}.${format === 'jpeg' ? 'jpg' : format}`,
-          image
-        );
+        if (image === 'data:,') {
+          const exportImageFailureMessage = await this.utilsService.translate(
+            'TOASTS.ERRORS.EXPORT_IMAGE_ERROR'
+          );
+          this.toastrService.error(exportImageFailureMessage);
+          return { success: false };
+        } else {
+          UtilsService.downloadFile(
+            `${name}.${format === 'jpeg' ? 'jpg' : format}`,
+            image
+          );
 
-        return { success: true, size: image.length / 1024 };
+          return { success: true, size: image.length / 1024 };
+        }
       }
     }
   }
--- a/teammapper-frontend/src/assets/i18n/it.json
+++ b/teammapper-frontend/src/assets/i18n/it.json
@@ -166,7 +166,8 @@
     "NODE_COPIED": "Nodo copiato con successo!",
     "ERRORS": {
       "ROOT_NODE_COPIED": "Non è possibile copiare il nodo radice!",
-      "GENERIC_NODE_COPY_ERROR": "Si è verificato un errore durante il tentativo di copiare il nodo radice."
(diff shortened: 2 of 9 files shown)


teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
```diff
@@ -471,12 +471,20 @@ export class MmpService implements OnDestroy {
       case 'png': {
         const image = await this.exportAsImage(format);
 
-        UtilsService.downloadFile(
-          `${name}.${format === 'jpeg' ? 'jpg' : format}`,
-          image
-        );
+        if (image === 'data:,') {
```
JannikStreek, line 474: cool thanks. Refactoring suggestions:

1) Make a constant EMPTY_IMAGE_DATA where you put the "data:..."
2) Extract method refactoring would be nice here to make this already big method smaller.

teammapper-frontend/src/app/core/services/mmp/mmp.service.ts
```diff
@@ -563,4 +494,101 @@ export class MmpService implements OnDestroy {
...
+      );
+
+      return { success: true, size: image.length / 1024 };
+    }
+  }
+
+  /**
+   * Export map to PDF
+   */
+  private async exportToPDF(format: string, name: string) {
```
JannikStreek, line 523: fyi: As far as I know its good practice to always define a typescript return type to functions. Yes, the return value can be inferred from the return value. But to make sure the return matches the writers specific , it's good to add it.

fyi2: pr ist still in draft mode

sorenjohanson: Return type can be done. It's still in draft mode because I want to implement the node boundary in this PR as well, but it's _complicated_.

sorenjohanson:
Adding a force boundary to prevent nodes from going outside of bounds (and preventing export) will be fixed in a separate PR.

JannikStreek:
looks good, is it possible to make the error more concrete in this case? e.g. threshold for mindmap size reached ?

teammapper-frontend/src/assets/i18n/it.json
```diff
@@ -166,7 +166,8 @@
     "NODE_COPIED": "Nodo copiato con successo!",
     "ERRORS": {
       "ROOT_NODE_COPIED": "Non è possibile copiare il nodo radice!",
-      "GENERIC_NODE_COPY_ERROR": "Si è verificato un errore durante il tentativo di copiare il nodo radice."
+      "GENERIC_NODE_COPY_ERROR": "Si è verificato un errore durante il tentativo di copiare il nodo radice.",
```
JannikStreek, line 169: as in the other PR, suggestion regarding naming of translations: `NODE_COPY_GENERIC_ERROR` or something

sorenjohanson:
> looks good, is it possible to make the error more concrete in this case? e.g. threshold for mindmap size reached ?

Don't see why not, I kept it generic since we don't know if any other errors might occur down the line (as all it checks is if the image data is empty, that could happen for a number of reasons).

teammapper-frontend/src/assets/i18n/it.json
sorenjohanson: This has been done in a separate PR: https://github.com/b310-digital/teammapper/pull/330

PR #306, opened 2024-07-02
Display an error when map cannot be found

### LINKED ISSUE #305: Display an error when map cannot be found
**Proposed feature**
When trying to access a non-existent map, whether `/map/abcd` or with a proper (non-existent) UUID, an error is displayed in the console:

`TypeError: Cannot read properties of null (reading 'uuid')`

We should instead display a toast that the map couldn't be found.

sorenjohanson:
closes #305

### DIFF
--- a/teammapper-frontend/angular.json
+++ b/teammapper-frontend/angular.json
@@ -80,15 +80,15 @@
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
             "proxyConfig": "src/proxy.conf.json",
-            "buildTarget": "teammapper:build",
-            "disableHostCheck": true
+            "buildTarget": "teammapper:build"
           },
           "configurations": {
             "production": {
               "buildTarget": "teammapper:build:production"
             },
             "development": {
-              "buildTarget": "teammapper:build:development"
+              "buildTarget": "teammapper:build:development",
+              "disableHostCheck": true
             }
           },
           "defaultConfiguration": "development"
--- a/teammapper-frontend/angular.json
+++ b/teammapper-frontend/angular.json
@@ -80,7 +80,8 @@
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
             "proxyConfig": "src/proxy.conf.json",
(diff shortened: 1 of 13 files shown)


teammapper-frontend/angular.json
```diff
@@ -87,7 +87,8 @@
               "buildTarget": "teammapper:build:production"
             },
             "development": {
-              "buildTarget": "teammapper:build:development"
+              "buildTarget": "teammapper:build:development",
```
JannikStreek, line 90: disablehostcheck should be removed I guess?

teammapper-frontend/src/assets/i18n/de.json
```diff
@@ -167,7 +167,8 @@
     "NODE_COPIED": "Knoten erfolgreich kopiert!",
     "ERRORS": {
       "ROOT_NODE_COPIED": "Es ist nicht möglich, den Hauptknoten zu kopieren!",
-      "GENERIC_NODE_COPY_ERROR": "Beim Kopieren des Knotens ist ein Fehler aufgetreten."
+      "GENERIC_NODE_COPY_ERROR": "Beim Kopieren des Knotens ist ein Fehler aufgetreten.",
```
JannikStreek, line 170: I would suggest to use a clearer prefixing of translations, e.g.:

`NODE_COPY_GENERIC_ERROR` start with the most important information first and not last. ERROR at the end here is fine or could be removed as you already have it under this key. Otherwise I would put it first, e.g. `ERROR_NODE_COPY_GENERIC`

teammapper-frontend/angular.json
sorenjohanson: No this is actually the correct place to put it, since this won't affect production and we won't have a billion angular.json changes in every PR just because of me :D

teammapper-frontend/src/assets/i18n/de.json
sorenjohanson: This is now in a separate PR: https://github.com/b310-digital/teammapper/pull/330

PR #325, opened 2024-07-05
Hide/show child nodes of a branch

### LINKED ISSUE #105: Hide/Show child nodes of a branch
**Proposed feature**
Currently it is not possible to hide branches of trees which makes big mind maps complicated to read. Therefore, a possibility to hide branches would be nice.

- [ ] Add a new toolbar icon to hide a branch, search for a good icon which changes depending on hidden/visible state.
- [ ] It should set the node state to invisible
- [ ] Invisible nodes shouldn't be shown in the map
- [ ] Check node data model if a new field is needed
- [ ] It might be helpful to show a small symbol next to a node if the branch is set to invisible. Otherwise one might overlook that.

sorenjohanson:
closes #105


sorenjohanson:
Do we want this to be a client-side only feature or server-side as well? Right now, there's still a bug left that the hidden state is not updated across all clients immediately (though I'd assume it calls map.draw.update() whenever something changes), but nodes are correctly hidden if the client updates the page.

Regarding an icon: Somehow very tricky to implement correctly and synchronize across all clients simultaneously.

JannikStreek:
> Do we want this to be a client-side only feature or server-side as well? Right now, there's still a bug left that the hidden state is not updated across all clients immediately (though I'd assume it calls map.draw.update() whenever something changes), but nodes are correctly hidden if the client updates the page.
> 
> Regarding an icon: Somehow very tricky to implement correctly and synchronize across all clients simultaneously.

good question, I would say as it only affects visibility it should only be on the client side? otherwise people might confuse this for that a user deleted a branch.

sorenjohanson:
> good question, I would say as it only affects visibility it should only be on the client side? otherwise people might confuse this for that a user deleted a branch.

Does it make sense to keep the changes to the data model then? Since those changes are synced server-side, inevitably leading new clients to consider selected nodes hidden as well.

JannikStreek:
> > good question, I would say as it only affects visibility it should only be on the client side? otherwise people might confuse this for that a user deleted a branch.
> 
> Does it make sense to keep the changes to the data model then? Since those changes are synced server-side, inevitably leading new clients to consider selected nodes hidden as well.

Either we sync it across clients so everybody is watching the same and we have it in our data model - or we just store that property on the client side without adding it to the data model. I think I would prefer the latter one.

sorenjohanson:
Works up to 99% percent, but there's still a weird and oddly specific bug where if two clients hide the same branches at the same time (so branch A and B), client A will unhide branch A when clicking on branch B.

Also missing: Changing the eye icon in the toolbar based on whether or not any child nodes are hidden.

JannikStreek:
> Works up to 99% percent, but there's still a weird and oddly specific bug where if two clients hide the same branches at the same time (so branch A and B), client A will unhide branch A when clicking on branch B.
> 
> Also missing: Changing the eye icon in the toolbar based on whether or not any child nodes are hidden.

i dont really understand this problem if hiding/showing is a local operation 🤔

sorenjohanson:
> > Works up to 99% percent, but there's still a weird and oddly specific bug where if two clients hide the same branches at the same time (so branch A and B), client A will unhide branch A when clicking on branch B.
> > Also missing: Changing the eye icon in the toolbar based on whether or not any child nodes are hidden.
> 
> i dont really understand this problem if hiding/showing is a local operation 🤔

It's not really consistently reproducible, but it's there - I'm trying to figure out if it maybe has something to do with how nodes selected by other users are rendered (maybe it affects the selectedNode variable in nodes.ts?). It's purely a client side issue though, most probably in how D3 renders the nodes, especially when other clients are involved.

sorenjohanson:
Think I found the issue: The NodeUpdated event gets called for Client B whenever Client A hides certain nodes. However, if Client B has already hidden the same set of nodes, the "hidden" values gets overwritten with null, since `UtilsService.get(newNode, propertyPath)` doesn't return a proper value. That means Client B needs to click _twice_ to unhide already hidden nodes, since it's set from null to true to false.

This can also lead to very funny issues when Client A hides a child, Client B hides every node except the root node and Client A then decides to _also_ hide the root node, because this then hides every node _except_ the originally hidden child.

sorenjohanson:
> a web socket update. An mmp event is fine as it's the local map that was updated. That should, however, not trigger a web socket

notifyWithEvent is true by default, forgot to change that to false.

Good point about the local data model - I'll have to test that. Hiding nodes is much more stable now, though I do have odd one-off cases where the parent nodes get hidden whilst children are still visible. I'll set the PR to open for review soon and then we can both test and iron out any kinks.

sorenjohanson:
Updating the data model works fine even when nodes are hidden. One thing that's still missing is some visible way to tell nodes are hidden, either by updating the icon in the toolbar or by attaching an SVG icon inside the drawing area to nodes.

JannikStreek:
> Updating the data model works fine even when nodes are hidden. One thing that's still missing is some visible way to tell nodes are hidden, either by updating the icon in the toolbar or by attaching an SVG icon inside the drawing area to nodes.

a small svg icon somewhere in the node sounds good 👍 the toolbar should also change (thats already done for add/remove image). But I think an additional item inside the node makes also sense as otherwise one might forget it.

teammapper-backend/src/map/types.ts
```diff
@@ -46,6 +46,7 @@ export interface IMmpClientNodeBasics {
 export interface IMmpClientNode extends IMmpClientNodeBasics {
   coordinates: IMmpClientCoordinates
   detached: boolean
+  hidden: boolean
```
JannikStreek, line 49: not needed anymore, if we only do this on the client side?

sorenjohanson: Would have to check, but if IMmpClientNode is only used for client/server comms then it's good to remove. I obviously need the hidden value on client side to actually hide nodes, but I think we primarily use UserNodeProperties for that.

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -55,13 +55,23 @@ export default class Draw {
      * Update the dom of the map with the (new) nodes.
      */
     public update() {
-        const nodes = this.map.nodes.getNodes(),
-            dom = {
-                nodes: this.map.dom.g.selectAll('.' + this.map.id + '_node').data(nodes),
-                branches: this.map.dom.g.selectAll('.' + this.map.id + '_branch').data(nodes.slice(1))
-            }
+        let nodes = this.map.nodes.getNodes().filter(x => !x.hidden)
```
JannikStreek, line 58: lets rename x to node

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -203,6 +204,35 @@ export default class Nodes {
         }
     }
 
+    /**
+     * Toggle (hide/show) all child nodes of selected node
+     */
+    public toggleBranch = () => {
```
JannikStreek, line 210: toggleBranchVisiblility

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -55,13 +55,23 @@ export default class Draw {
...
+
+        const dom = {
+            nodes: this.map.dom.g.selectAll('.' + this.map.id + '_node').data(nodes, (d) => d.id),
+            branches: this.map.dom.g.selectAll('.' + this.map.id + '_branch').data(nodes.slice(1), (d) => d.id)
+        }
         let tapedTwice = false
 
+        dom.nodes.each((node: Node) => {
```
JannikStreek, line 66: wouldnt it be easier to set the hidden field directly when the button is pressed for all nodes, and not inside this render method?

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -288,6 +298,35 @@ export default class Draw {
         }
     }
 
+    /**
+     * Set a hidden eye icon if child nodes are hidden.
+     * @param {Node} node
+     */
+    public setHiddenChildrenIcon(node: Node) {
```
JannikStreek, line 305: maybe we can extract parts of the method as its similar to the link icon?

teammapper-frontend/src/app/modules/application/components/toolbar/toolbar.component.ts
```diff
@@ -29,6 +29,15 @@ export class ToolbarComponent {
       );
   }
 
+  get hasHiddenNodes() {
+    return this.mmpService.nodeChildren()?.filter(x => x.hidden).length > 0
```
JannikStreek, line 33: x => node

teammapper-frontend/mmp/src/map/handlers/draw.ts
sorenjohanson: Don't think so personally, there's just a little bit of overlap and it's two very different things at the end of the day (setting a link when clicked vs automatically setting a very specific icon when conditions apply)

teammapper-frontend/mmp/src/map/handlers/draw.ts
sorenjohanson: Not sure what you mean by this? There's practically no difference whether you actually set a field on the node itself or do it like this, as you'd still have to call both methods (`setHiddenChildrenIcon` and `removeHiddenChildrenIcon`) separately whenever you add/remove the icon to make it work reliably, at least in my testing.

You also can't avoid calling `dom.nodes.each` as `outer.each` (which is based on the `enter` method of D3.js) only seems to include nodes that were changed, ie those that were hidden, whereas we explicitly want the node that wasn't hidden.

teammapper-frontend/mmp/src/map/handlers/draw.ts
JannikStreek: i am only referring to the icon, not to the logic. e.g. creating an element, setting the icon name, setting the position according to parameters etc. Thats not just one line. BTW there is I think a copy paste error: `domText.classList.add('link-text')` ;-)

sorenjohanson: Yeah `link-text` was unnecessary, removed.

JannikStreek:
Looks good. Two bugs that need more checking:

- Undo/Redo is bugged - I got very strange results with it with nodes flying around in space without being attached. If possible I wouldn't consider visibility changes as un/redoable. But thats just on first thought.
- Hiding nodes and then exporting the image looks like this:
Bildschirmfoto 2024-07-29 um 16 18 03

Not sure what the preferred behavior for this feature is on exports. Either completely hide it (like it is now) or just show it - but without this visibility note thingy. The same is btw true for the link, which just shows "link" - but thats a separate issue.

sorenjohanson:
> Not sure what the preferred behavior for this feature is on exports. Either completely hide it (like it is now) or just show it - but without this visibility note thingy. The same also happens for attached links, which just show "link" - but thats a separate issue.

Re this: I think keeping notes invisible when exporting is the preferred option. However, it seems like it doesn't work with Material Icons, so I'm guessing removing the link text ("visibility_off") is the better option.

sorenjohanson:
I'll take a look at undo/redo.

sorenjohanson:
I don't think there's much we can do about undo/redo without a rework. The issue is that if we hide two nodes, that's two steps in the snapshot, with the farthest node being the first step in the snapshot to be undone. Similar weird behaviour can be observed when creating a node with text and undoing it: Creating the node is one step, with the text being a separate step, so it takes two presses of undo to delete a single node.

sorenjohanson:
> Not sure what the preferred behavior for this feature is on exports. Either completely hide it (like it is now) or just show it - but without this visibility note thingy. The same also happens for attached links, which just show "link" - but thats a separate issue.

I've removed all text on export now, so neither "link" nor "visibility_off" should appear on export.

JannikStreek:
There are still bugs: If another user adds nodes to a hidden parent node, it gets weird:

Bildschirmfoto 2024-08-10 um 14 08 34
Bildschirmfoto 2024-08-10 um 14 08 29

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -235,13 +235,20 @@ export default class Nodes {
...
+
                     if (!x.parent.hidden && x.hidden) {
                         this.updateNode('hidden', false, false, false, false, x.id)
                     }
                 })
             }
 
+            // Lengthy but definitive check to see if we have any hidden nodes after toggling
+            if (this.map.nodes.nodeChildren(this.selectedNode.id)?.filter(x => x.hidden).length > 0) {
+                this.selectedNode.hasHiddenChildNodes = true
```
JannikStreek, line 247: can't this be set directly, when the button is clicked (hide children)? Do we have to calculate this dynamically?

JannikStreek: As I see it, this property makes sense. So we have to new fields:

- hidden (on the descendants) 
- hiddenBranches (on the parent) 

So I would set them both upon button press?

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -55,16 +55,26 @@ export default class Draw {
      * Update the dom of the map with the (new) nodes.
      */
     public update() {
-        const nodes = this.map.nodes.getNodes(),
-            dom = {
-                nodes: this.map.dom.g.selectAll('.' + this.map.id + '_node').data(nodes),
-                branches: this.map.dom.g.selectAll('.' + this.map.id + '_branch').data(nodes.slice(1))
-            }
+        let nodes = this.map.nodes.getNodes()
```
JannikStreek, line 58: why let here? do you need to modify this ?

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -265,29 +282,59 @@ export default class Draw {
      */
     public setLink(node: Node) {
         let domLink = node.getLinkDOM()
+        const domText = document.createElementNS('http://www.w3.org/2000/svg', 'text')
```
JannikStreek, line 285: why was this change needed? On first glance, if the link Dom is already present, it makes no sense to draw on top of it?

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -399,34 +446,46 @@ export default class Draw {
         }
     }
 
+    /**
+     * Check if given node has hidden children and render the eye icon if so
+     * @param {Node} node
+     */
+    private checkForHiddenChildren(node: Node) {
```
JannikStreek, line 453: this method does not only check, but actually hides or even removes Dom icons. That should be reflected in the name, as its otherwise confusing

teammapper-frontend/mmp/src/map/handlers/history.ts
```diff
@@ -54,17 +54,40 @@ export default class History {
 
             this.save()
 
-            if(notifyWithEvent) this.map.events.call(Event.create, this.map.dom)
+            if (notifyWithEvent) this.map.events.call(Event.create, this.map.dom)
         } else if (this.checkSnapshotStructure(snapshot)) {
             const previousData = this.map.export.asJSON()
 
+            // Find all nodes where we've set hasHiddenChildNodes in the previous map
```
JannikStreek, line 61: maybe apply refactoring: extract method, as this is a lot of logic only dedicating to the hiding of nodes.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -889,6 +943,24 @@ export default class Nodes {
...
+     * @param {boolean} hidden
+     * @returns {boolean}
+     */
+    private updateNodeHidden = (node: Node, hidden: boolean) => {
+        if (hidden && typeof hidden !== 'boolean') {
+            Log.error('The hidden value must be boolean', 'type')
+        }
+
+        if (node.hidden !== hidden) {
+            node.hidden = hidden;
```
JannikStreek, line 958: please check to return a consistent return type. here, you either return false or you assign a variable and return undefined (which is also false).

teammapper-frontend/mmp/src/map/handlers/draw.ts
sorenjohanson: Good catch, must've been a mistake from when I separated out nodes and dom variable creation.

teammapper-frontend/mmp/src/map/handlers/draw.ts
sorenjohanson: Was testing to see if it made a difference in rendering, I should be able to revert the changes to the method since the issue was with the querySelector and not the setLink() method.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: All the other methods also either assign a variable or return false, so I was just orienting myself on those. Maybe a general refactoring idea?

JannikStreek: imho only because the other methods follow a bad practice, this doesn't mean we have to continue this way. The return value is useless if it's always false? But yes, the other methods could be refactored in a separate refactoring.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
JannikStreek: good point, maybe you can add this reason as a comment? that there is not only the button that triggers it and thats the reason you have to check here.

sorenjohanson: Good idea, I've added some comments.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: Yeah, I've refactored this one to return undefined, I'll open an issue for the other ones.

PR #337, opened 2024-07-17
History bugs when copying map

### LINKED ISSUE #336: History bugs when copying map
**Describe the bug**
When copying a map, it successfully redirects to the new map, but going back doesn't seem to work properly.

We should look at using window.location.replace() instead as we did before to prevent these issues from appearing.

sorenjohanson:
closes #336


teammapper-frontend/src/app/guards/toast.guard.ts
```diff
@@ -29,8 +29,9 @@ export class ToastGuard implements CanActivate {
 
       // This preserves both map UUID and the all important fragment whilst deleting anything toast-related.
       const urlTree = this.router.parseUrl(state.url);
+      console.log(urlTree);
```
JannikStreek, line 32: remove log ;)

PR #348, opened 2024-07-22
Reload map when navigation is coming from back/forward

### LINKED ISSUE #347: Reload map when navigation is coming from back/forward
**Proposed feature**
The map needs to be reloaded when someone is navigating from in a different page back to the map. This is especially important, if multiple maps were opened and navigation is used to switch between those. Otherwise it looks like you have opened the right map but in the background you are indeed editing a different one.

Note: Using the "new map button" top left in navigation, the process already works without bugs. It is primarily the copy map button that is affected by this behavior.

Update: So far, this seems to be only a problem for safari+chrome on Mac

sorenjohanson:
closes #347


JannikStreek:
that works in chrome on my mac, but somehow not in safari

teammapper-frontend/src/app/root.component.ts
```diff
@@ -40,6 +40,15 @@ export class RootComponent implements OnInit {
         replaceUrl: true,
       });
     }
+
+    // Fix for #347: Force a reload of mindmaps, even when they are persisted in bfcache
```
JannikStreek, line 44: please make this comment a bit more descriptive. It's already obvious that a reload happens, the question is why ;) e.g. "Force reload to prevent broken sync state on Mac safari where the URL and the internal state don't match"

sorenjohanson:
> that works in chrome on my mac, but somehow not in safari

Can't reproduce this on iPad Pro using Safari, so guess I'll have to buy a Macbook to fix this bug. :D 

I changed it to `window.location = window.location`, maybe that helps? Incredibly unlikely, but this is already a niche bug.

PR #356, opened 2024-07-30
Update from Angular v17.x to v18.x

### LINKED ISSUE #355: Angular Update from v17.x to v18.x
**Proposed feature**
Angular needs to be updated from v17.x to v18.x, see dependabot: https://github.com/b310-digital/teammapper/pull/345

sorenjohanson:
closes #355 

Also closes #350, #345, #317


sorenjohanson:
This bump adds an unexpected bug that causes mindmaps to no longer be found when navigating back/forward, or by other clients, or when changing settings. No cause found yet.

sorenjohanson:
`HttpClient` is deprecated and has been replaced with `provideHttpClient`, maybe this is the issue. It seems to fail somewhere between the frontend making the request with fetch and the backend, but the backend returns the map just fine.

teammapper-backend/src/map/controllers/maps.controller.ts
```diff
@@ -34,6 +34,7 @@ export default class MapsController {
     @Param('id') mapId: string,
     @Body() body: IMmpClientDeleteRequest
   ): Promise<void> {
+    console.log("Deleting: ", body)
```
JannikStreek, line 37: remove :)

teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -56,7 +56,10 @@ export class MapsGateway implements OnGatewayDisconnect {
     @ConnectedSocket() client: Socket,
     @MessageBody() request: IMmpClientJoinRequest
   ): Promise<IMmpClientMap> {
-    if (!(await this.mapsService.findMap(request.mapId)))
+    console.log("Joining: ", request)
```
JannikStreek, line 59: same, remove please :)

teammapper-frontend/package.json
```diff
@@ -73,15 +73,15 @@
...
+    "@angular-devkit/core": "^18.1.2",
+    "@angular-devkit/schematics": "^18.1.2",
+    "@angular-eslint/builder": "18.1.0",
+    "@angular-eslint/eslint-plugin": "18.1.0",
+    "@angular-eslint/eslint-plugin-template": "18.1.0",
+    "@angular-eslint/schematics": "18.1.0",
+    "@angular-eslint/template-parser": "18.1.0",
+    "@angular/language-service": "^18.1.2",
     "@compodoc/compodoc": "^1.1.25",
     "@schematics/angular": "^17.3.2",
```
JannikStreek, line 86: is there also a v18 available?

teammapper-frontend/src/app/modules/about/components/jumbotron/jumbotron.component.scss
```diff
@@ -2,7 +2,7 @@
 @import 'src/theme.scss';
 
 section.jumbotron {
-  background-color: mat.get-color-from-palette($app-primary);
+  background-color: mat.m2-get-color-from-palette($app-primary);
```
JannikStreek, line 5: why do we have to use the older version here? Guess the default is v3 now? Was the design broken? We can stay with version 2 but then we should create a new issue about updating to material v3. But lets only consider this if there were problems.

sorenjohanson: This was done by the auto-updater, because `mat.get-color-from-palette` is not a valid function (only if prefixed by `m2-`, same goes for variables).

teammapper-frontend/package.json
sorenjohanson: Yup, which will already be pulled in by @angular/cli, but I've updated the version to ^18.1.4 anyway.

JannikStreek:
Have you checked for all points in the update guide?  => https://angular.dev/update-guide?v=17.0-18.0&l=2 cli should take care of most of them I guess. But not necessarily  all.

teammapper-frontend/src/app/modules/about/components/jumbotron/jumbotron.component.scss
JannikStreek: I created a new issue for this #367

JannikStreek:
looks good to me

PR #366, opened 2024-08-08
Add optional boolean argument to updateNode() to skip updating history

### LINKED ISSUE #354: Revamp snapshot history to include multiple steps
**Proposed feature**
Right now, undo/redo works on simply going back/forward one step at a time. This is awkward because even something as simple as creating one node with text involves two steps (creating the node and then the text), so removing a node effectively requires pressing undo twice.

Instead, various functions like copy paste, creating a new node with text and also hiding nodes should be summarized in one snapshot, which can then be cleanly undone or redone without steps inbetween.

sorenjohanson:
closes #354

### DIFF
--- a/teammapper-frontend/mmp/src/map/handlers/nodes.ts
+++ b/teammapper-frontend/mmp/src/map/handlers/nodes.ts
@@ -82,7 +82,7 @@ export default class Nodes {
      * @param {string} parentId
      * @param {string} overwriteId
      */
-    public addNode = (userProperties?: UserNodeProperties, notifyWithEvent: boolean = true, parentId?: string, overwriteId?: string): Node => {
+    public addNode = (userProperties?: UserNodeProperties, notifyWithEvent: boolean = true, updateHistory: boolean = true, parentId?: string, overwriteId?: string): Node => {
         const parentNode: Node = userProperties.detached ? null :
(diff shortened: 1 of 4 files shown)


teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -287,7 +287,9 @@ export default class Nodes {
                 Log.error('The property does not exist')
         }
         if (graphic === false && updated !== false) {
```
JannikStreek, line 289: please lets try to avoid so much if nesting. Example how it could be flatter:

```
  if (graphic === false && updated !== false && updateHistory) {
  this.map.history.save()
}

if (graphic === false && updated !== false && notifyWithEvent) {
  this.map.events.call(Event.nodeUpdate, node.dom, { nodeProperties: this.getNodeProperties(node), changedProperty: property, previousValue })
  }
```

Even better, create new methods that check for the conditions (remember to put these anonymous functions above the updateNode method or make real functions of of them, otherwise it won't work):

```
const shallUpdateWithEvent = () => graphic === false && updated !== false && notifyWithEvent;
const shallUpdateWithHistory= () => graphic === false && updated !== false && updateHistory
```

JannikStreek:
Cool, I like this new option. Please make sure, that a map with multiple users still works as expected with undo/redo etc.

PR #390, opened 2024-08-15
Upgrade @types/uuid package to dependency

### LINKED ISSUE #387: Upgrade @types/uuid package to dependency
**Proposed feature**
For build purposes we should move the @types/uuid package to be a dependency and not a devDependency. We should also double check the version we're using (9 or 10).

sorenjohanson:
closes #387


JannikStreek:
I think some part of nests uses version 9 of uuid. Maybe we can upgrade this package to have it consistent?

JannikStreek:
decided to stay with v10

PR #392, opened 2024-08-15
Refactor methods in nodes.ts to return consistent values

### LINKED ISSUE #381: Refactor methods in nodes.ts to return consistent value
**Proposed feature**
Currently, all methods found in `updateNode` return inconsistent values: They either update a variable or return false (so the return is either void or boolean). They should be updated to return `void` consistently, so `undefined` instead of `false` if variables are not updated.

sorenjohanson:
closes #381

### DIFF
--- a/teammapper-frontend/mmp/src/map/handlers/nodes.ts
+++ b/teammapper-frontend/mmp/src/map/handlers/nodes.ts
@@ -724,8 +724,8 @@ export default class Nodes {
 
             if (graphic === false) {
                 node.name = name
-                return true
             }
+            return true
         } else {
             return false
         }
@@ -784,8 +784,8 @@ export default class Nodes {
 
             if (graphic === false) {
                 node.colors.background = sanitizedColor
-                return true
             }
+            return true
         } else {
             return false
         }
@@ -810,8 +810,8 @@ export default class Nodes {
 
             if (graphic === false) {
                 node.colors.name = sanitizedColor
-                return true
             }
+            return true
         } else {
             return false
         }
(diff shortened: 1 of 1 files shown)


teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -806,6 +810,7 @@ export default class Nodes {
 
             if (graphic === false) {
                 node.colors.name = sanitizedColor
+                return true
```
JannikStreek, line 813: if graphic is true and `node.colors.name !== color`, then the return value is still undefined.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -806,6 +810,7 @@ export default class Nodes {
 
             if (graphic === false) {
```
JannikStreek, line 811: we could refactor this to `if (graphic)` as long as undefined isn't a special case which is different to false.

sorenjohanson: This might be unnecessary considering https://github.com/b310-digital/teammapper/pull/369, where we remove `graphic` entirely.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: Good point, I've moved the `return true` outside of the `if (graphic)` since the node has already been updated anyway.

PR #404, opened 2024-08-21
Add createdAt date time to maps

### LINKED ISSUE #403: Add createdAt date time to maps
**Proposed feature**
We want to show the last maps on the start page with the time at which they were created. However, when a map is created in the backend, the response includes a ServerMap object which includes the following data:

```
export interface IMmpClientMap {
  uuid: string
  lastModified: Date
  deleteAfterDays: number
  deletedAt: Date
  data: IMmpClientNode[]
  options: IMmpClientMapOptions
}
```

This should be enhanced by the created time of the server map. However, I don't think this field is currently available in the database.

- [ ] Create a new `createdAt`field in the database for the map table
- [ ] Export the `createdAt` field on map creation in the endpoint
- [ ] Add the new field to the frontend types and save the field in the storage service within `prepareNewMap` call from map-sync Service

sorenjohanson:
closes #403

### DIFF
--- a/teammapper-backend/src/map/entities/mmpMap.entity.ts
+++ b/teammapper-backend/src/map/entities/mmpMap.entity.ts
@@ -36,4 +36,7 @@ export class MmpMap {
   })
   /* eslint-enable @typescript-eslint/no-unused-vars */
   nodes: MmpNode[]
+
+  @Column({ type: 'timestamptz', default: () => 'now()' })
+  createdAt: Date
 }
--- a/teammapper-backend/src/map/entities/mmpMap.entity.ts
+++ b/teammapper-backend/src/map/entities/mmpMap.entity.ts
@@ -37,6 +37,6 @@ export class MmpMap {
   /* eslint-enable @typescript-eslint/no-unused-vars */
   nodes: MmpNode[]
 
-  @Column({ type: 'timestamptz', default: () => 'now()' })
+  @Column({ type: 'timestamptz' })
   createdAt: Date
 }
--- a/teammapper-backend/src/map/entities/mmpMap.entity.ts
+++ b/teammapper-backend/src/map/entities/mmpMap.entity.ts
@@ -37,6 +37,6 @@ export class MmpMap {
   /* eslint-enable @typescript-eslint/no-unused-vars */
   nodes: MmpNode[]
 
-  @Column({ type: 'timestamptz' })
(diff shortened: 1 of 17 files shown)


teammapper-backend/src/map/entities/mmpMap.entity.ts
```diff
@@ -36,4 +36,7 @@ export class MmpMap {
   })
   /* eslint-enable @typescript-eslint/no-unused-vars */
   nodes: MmpNode[]
+
+  @Column({ type: 'timestamptz', default: () => 'now()' })
```
JannikStreek, line 40: the default ist not really fitting here. if possible use the modifiedAt as default or leave it null

teammapper-backend/src/map/entities/mmpNode.entity.ts
```diff
@@ -102,4 +102,7 @@ export class MmpNode {
 
   @Column({ type: 'timestamptz', nullable: true, default: () => 'now()' })
   lastModified: Date
+
+  @Column({ type: 'timestamptz', default: () => 'now()' })
```
JannikStreek, line 106: same

JannikStreek:
Please also provide the field to the frontend, currently it's not exposed/returned. Please see issue #403 for details.

teammapper-backend/src/map/entities/mmpMap.entity.ts
sorenjohanson: I've removed the default here entirely, though I'm guessing that we'll create new maps using `createdAt: new Date()`, so I'm not sure how much of a practical difference that'll make.

teammapper-backend/src/map/entities/mmpNode.entity.ts
sorenjohanson: Same as above.

teammapper-backend/src/map/entities/mmpMap.entity.ts
JannikStreek: my primarily concern would be existing maps, where this field should be empty. I think adding a new column with a default value would populate also existing old entries which is not wanted.

sorenjohanson: Added defaults now in a secondary migration, so existing maps get a null entry in createdAt.

PR #437, opened 2024-09-04
Add back zoom buttons and shortcuts

### LINKED ISSUE #436: Add back zoom buttons and shortcuts
**Proposed feature**
We removed the zoom buttons in https://github.com/b310-digital/teammapper/commit/454c31c88632b9df00e6e3c2a8020382dc599968 because it seemed at that time that zooming with the touchpad or buttons (CMD+PLUS) would be enough. Unfortunately, this doesn't work as good as thought. Often, the zoom also makes the menu smaller/bigger. 

Therefore, we need the zoom buttons back which were removed in the given commit.

sorenjohanson:
closes #436


JannikStreek:
lint is failing, but otherwise looking good

PR #456, opened 2024-09-12
Fix warning non-passive event handler

### LINKED ISSUE #454: Fix warning: Added non-passive event listener to a scroll-blocking 'touchstart' event
**Describe the bug**
Every time when a node is added, I see this warning in the console
```
draw.ts:89 [Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952
```

Task: Investigate why its there and preferably fix it as well.

sorenjohanson:
closes #454

### DIFF
--- a/teammapper-frontend/mmp/src/map/handlers/draw.ts
+++ b/teammapper-frontend/mmp/src/map/handlers/draw.ts
@@ -106,7 +106,7 @@ export default class Draw {
                 }
 
                 this.enableNodeNameEditing(node)
-            })
+            }, { passive: false })
         if (this.map.options.drag === true) {
             outer.call(this.map.drag.getDragBehavior())
         } else {


sorenjohanson:
The change to the API is documented here: https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md

It mostly talks about adding `{ passive: true }` to prevent the handler from blocking scrolling (which it does by default), but to remove the warning you have to explictly tell the browser that the event handler is not passive by using `{ passive: false }`. A `wheel` or any `touch*` event handler is only passive if it never calls `event.preventDefault()`.

There's https://github.com/zzarcon/default-passive-events if we ever have a situation with a lot of `wheel` our `touch*` event handlers, because `passive` is never set by default, either to true or false, so it always has to be set explictly.

PR #467, opened 2024-09-16
Bug: Maps with zero nodes

### LINKED ISSUE #455: bug: maps with zero nodes
On our servers there are some maps which do not contain any nodes. That can't really be. The only reason which comes to my mind is a failed json import. I think we delete all nodes before adding the new nodes. If the import fails. the map might end up with zero nodes. 

_Originally posted by @JannikStreek in https://github.com/b310-digital/teammapper/issues/296#issuecomment-2346110509_

sorenjohanson:
reference to  #455


teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -176,14 +180,26 @@ export class MapsGateway implements OnGatewayDisconnect {
       return Promise.resolve(false)
 
     const mmpMap: IMmpClientMap = request.map
+
+    // Disconnect all clients temporarily
+    // Emit an event so clients can display a notification
+    this.server.to(mmpMap.uuid).emit('clientNotification', { clientId: client.id, message: 'TOASTS.WARNINGS.MAP_IMPORT_IN_PROGRESS', type: 'warning' })
+
+    this.clients.forEach((client) => client.leave(request.mapId))
```
JannikStreek, line 188: saving all clients in a global map could be problematic:

1) You are iterating over all available clients of all maps, but you are only caring about one map. First optimization would be definitely using a hash map and not a set. But I suspect that nestJS already provides mechanisms to get all clients in one room? Have you checked that?
2) It won't work on a multi container setup, but then again, I think nether the rest will. We would need a redis to share client status. => Could be actually a new issue, not but important right now.

teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -176,14 +177,30 @@ export class MapsGateway implements OnGatewayDisconnect {
...
+    // Emit an event so clients can display a notification
+    this.server.to(mmpMap.uuid).emit('clientNotification', { clientId: client.id, message: 'TOASTS.WARNINGS.MAP_IMPORT_IN_PROGRESS', type: 'warning' })
+
+    const sockets = await this.server.in(request.mapId).fetchSockets()
+    this.server.in(request.mapId).socketsLeave(request.mapId)
+
     await this.mapsService.updateMap(mmpMap)
 
+    // Reconnect clients once map is updated
+    sockets.forEach((socket: RemoteSocket<any, any>) => {
```
JannikStreek, line 191: hope there are better types available than any any? ;-)

sorenjohanson: Unfortunately not, because afaik it returns `RemoteSocket<DefaultEventMaps, any>`, but `socket.io` doesn't export `DefaultEventMaps`. Haven't found anything in the docs regarding `RemoteSocket` either, but I'll take another look.

JannikStreek: the other question is also if you need to explicitly specify the type as usually it should be inferred from the list of `fetchSockets()`. So maybe just remove the type definition. If the library doesn't specify more its ok but I don't like having `any` in the source code.

sorenjohanson: I changed it so the type is inferred from usage.

PR #492, opened 2024-09-30
Add client-side rollback during snapshot import

### LINKED ISSUE #490: Add client-side rollback during snapshot import
**Proposed feature**
When importing a snapshot, there's currently no way to roll back on errors client-side. To make it more robust, a "clean" snapshot should be saved before deleting nodes and importing the new ones, allowing a rollback to that clean snapshot in case of any errors.

sorenjohanson:
closes #490

### DIFF
--- a/teammapper-frontend/mmp/src/map/handlers/history.ts
+++ b/teammapper-frontend/mmp/src/map/handlers/history.ts
@@ -64,9 +64,14 @@ export default class History {
 
             this.map.zoom.center('position', 0)
 
-            this.save()
-
-            if (notifyWithEvent) this.map.events.call(Event.create, this.map.dom, { previousMap: previousData })
+            // If the amount of nodes is == 0, automatically rollback to the last clean snapshot and display a toast
+            if (this.map.nodes.getNodes().length === 0) {
+                this.undo()
+                Log.error('There was an error importing the map; changes have been rolled back.')
+            } else {
+                this.save()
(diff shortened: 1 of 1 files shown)


teammapper-frontend/mmp/src/map/handlers/history.ts
```diff
@@ -120,8 +125,11 @@ export default class History {
      * Redraw the map with a new snapshot.
      * @param {MapSnapshot} snapshot
      */
-    private redraw(snapshot: MapSnapshot) {
-        this.map.nodes.clear()
+    private redraw(snapshot: MapSnapshot, backup: boolean = false) {
```
JannikStreek, line 128: so when is this called? currently its not? its always false, isn't it?

JannikStreek:
How can I test this locally?

teammapper-frontend/mmp/src/map/handlers/history.ts
sorenjohanson: I forgot to add `this.redraw(snapshot, true)` to the snapshot import. :D 

This is only called during a snapshot import right now.

sorenjohanson:
Apparently still buggy, so I reset this to draft for now.

sorenjohanson:
The rollback uses `redraw` directly now without relying on `save()`, this should allow for a more robust undo and prevent stale snapshots (which was an issue before).

PR #504, opened 2024-10-11
Update to use ESLint 9 and Prettier 3

### LINKED ISSUE #479: Update to use ESLint 9 and Prettier 3.x
**Proposed feature**
Right now, we're still on ESLint 8 and Prettier 2.x. We should eventually update to ESLint 9 and Prettier 3.x, but it'll require some work. Dependabot keeps suggesting to update `prettier`, `eslint` or `eslint-plugin-prettier`. `prettier` 3.x requires `eslint-plugin-prettier` version 5, which in turn requires `eslint` version 9, so all of these updates should be ignored if dependabot suggests them again.

sorenjohanson:
closes #479


JannikStreek:
sorry, I should have merged this before the others. can you solve this conflicts so we can merge this next?

PR #512, opened 2024-10-14
Add ESLint Stylistic to legacy config

### LINKED ISSUE #409: Add ESLint Stylistic to incorporate formatting rules
**Proposed feature**
With the upgrade to typescript-eslint v8, there's been some deprecations, namely removing old formatting/layout rules. Practically, we're only using `@typescript-eslint/member-delimiter-style`, which has now been moved to https://eslint.style/. We should look at incorporating this in the future to keep the formatting rule.

sorenjohanson:
closes #409 - separate PR for legacy config. #504 includes the same for the new type of configuration.


sorenjohanson:
> can you double check that the package.lock file really only contains the changes for the linter? This looks more like changes from another pr maybe?

I've deleted everything locally and re-ran npm install for both, it's updated some versions but otherwise not much changes.

PR #514, opened 2024-10-18
Rework undo/redo to use diff

### LINKED ISSUE #370: Undo in a multi user session is handled differently amongst clients
**Describe the bug**
In a multi user session: If client A cuts + pastes nodes and undos the paste, the whole pasting step is correctly reverted. If client A pastes and client B reverts, it only reverts one node of the pasted content.

**Expected behavior**
Undo should behave the same amongst clients, reverting all the pasted content.

### LINKED ISSUE #379: Rework undo/redo to use diff
**Proposed feature**
Currently, undo/redo will simply go back/forward one step and afterwards send the entire map over the wire. We should rework this to use diffs instead so we only hand the actual changes over to new clients.

sorenjohanson:
closes #379 and fixes #370


teammapper-backend/src/map/utils/tests/mapFactories.ts
```diff
@@ -20,6 +20,17 @@ export const createMmpMap = (overrides = {}): MmpMap => ({
     ...overrides
 })
 
+export const createMmpRootNode = (mapId: string, overrides = {}): Partial<MmpNode> => ({
```
JannikStreek, line 23: Why Partial? Imho a factory should return a valid object?

teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -167,6 +168,41 @@ export class MapsGateway implements OnGatewayDisconnect {
...
+    @MessageBody() request: IMmpClientUndoRedoRequest
+  ): Promise<boolean> {
+    if (!(await this.mapsService.findMap(request.mapId)))
+      return Promise.resolve(false)
+    if (!request.diff)
+      return Promise.resolve(false)
+
+    // Disconnect all clients temporarily
+    // Emit an event so clients can display a notification
+    this.server.to(request.mapId).emit('clientNotification', { clientId: client.id, message: 'TOASTS.WARNINGS.MAP_UPDATE_IN_PROGRESS', type: 'warning' })
```
JannikStreek, line 184: disconnecting all clients just because a change like a name is reverted feels a big drastic? For the case of importing a whole different map it makes sense, but just updating a name or coordinates of a node should not drop all connections of the clients.

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
```diff
@@ -566,16 +613,18 @@ export class MapSyncService implements OnDestroy {
       this.updateAttachedMap();
     });
 
-    this.mmpService.on('undo').subscribe(() => {
+    this.mmpService.on('undo').subscribe((diff?: any) => {
```
JannikStreek, line 616: always try to avoid any where possible

JannikStreek:
It would help me to understand your strategy / concept first. What exactly happens currently if I redo / undo something on the server and what data is sent? Why do you need new methods on the server to update the state? As the client only redo's small changes, why do we need to send a diff map?

teammapper-backend/src/map/utils/tests/mapFactories.ts
sorenjohanson: Since this method is only used for particular test cases, I chose Partial to avoid having to specify things like colorsBackground, colorsName etc - these attributes don't add anything to the test except a lot of extra lines of code.

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
sorenjohanson: Was a leftover from testing, types already existed so I just updated it now.

teammapper-backend/src/map/controllers/maps.gateway.ts
sorenjohanson: Clients no longer get disconnected when executing undo/redo, seems to work fine.

teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -167,6 +168,26 @@ export class MapsGateway implements OnGatewayDisconnect {
     return true
   }
 
+  @UseGuards(EditGuard)
+  @SubscribeMessage('undoRedoMapChanges')
+  async undoRedoMapChanges(
```
JannikStreek, line 173: I would suggest to call it differently. the first time I actually struggled to understand what "undoRedo" should actually mean. e.g.: "applyMapDiff" or something like that.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -183,6 +185,43 @@ export class MapsService {
...
+          const newNode = new MmpNode();
+          Object.assign(newNode, mergeClientNodeIntoMmpNode(clientNode, newNode));
+          newNode.nodeMapId = mapId;
+          await this.nodesRepository.save(newNode);
+        }
+      }
+    }
+  
+    if (diff.updated && typeof diff.updated === 'object') {
+      for (const key in diff.updated) {
```
JannikStreek, line 203: Suggestion: `Object.keys(diff.updated).map(...)` ?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -183,6 +185,43 @@ export class MapsService {
     return this.findMap(clientMap.uuid)
   }
 
+  async updateMapByDiff(mapId: string, diff: IMmpClientMapDiff) {
+    if (diff.added && typeof diff.added === 'object') {
```
JannikStreek, line 189: this can be done easier, I think. Just a quick sketch, feel free to improve it:

```typescript
const diffAddedCallback = () => {...}
const diffRemovedCallback = () => {...}
const diffUpdatedCallback = () => {...}

const callbacks = {
'added': diffAddedCallback,
...
}

['added', 'updated', 'removed'].each((key) => {
  if(diff[key] && diff[key] === 'object') { // testing for object isn't really best practice, aren't there better alternatives?
    callbacks[key]()
  }
})

```

teammapper-frontend/mmp/src/map/handlers/history.ts
```diff
@@ -26,6 +27,22 @@ export default class History {
         this.snapshots = []
     }
 
+    private switchDiffKeys(snapshotDiff: MapDiff) {
```
JannikStreek, line 30: some comments would be nice here 😂

JannikStreek:
This is an important improvement, thanks. Are old tests already covering this? You added a new endpoint and service methods. These should be tested I guess?

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
```diff
@@ -429,6 +436,47 @@ export class MapSyncService implements OnDestroy {
       this.mmpService.new(result.map.data, false);
     });
 
+    this.socket.on('mapChangesUndoRedo', (result: ResponseUndoRedoChanges) => {
```
JannikStreek, line 439: same refactoring suggestion I mentioned.

Also: Might be a good place to display a toast? "Change reverted" / "Change re-applied" / "Nodes pasted" something like that? What do you think? I think its helpful for the users to understand what is happening.

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
```diff
@@ -566,16 +614,18 @@ export class MapSyncService implements OnDestroy {
       this.updateAttachedMap();
     });
 
-    this.mmpService.on('undo').subscribe(() => {
+    this.mmpService.on('undo').subscribe((diff?: MapDiff) => {
       this.attachedNodeSubject.next(this.mmpService.selectNode());
+      // Updating the attached map is important because this persists changes after refresh - it's ok to keep this as a full update for now.
```
JannikStreek, line 619: I would remove this comment or at least remove the second part "its ok to..." as it does not help much the understanding

teammapper-backend/src/map/utils/tests/mapFactories.ts
JannikStreek: ok. what I do not like that much is this part:

```
const rootNode = createMmpRootNode(localMap.id) as MmpNode
```

So you are casting now as you only have a partial but you want to have a full one ;) . Either you merge with something to get the values, or you have to construct a complete object. But this doesn't look like best practice to me.

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
sorenjohanson: The question is if it would be too spammy to display toasts, esp. when you have like ~10 users working on the same map. The toasts are quite big and present, so your attention will always be drawn to them - I think quite unhelpful if all they're doing is undo/redo.

teammapper-backend/src/map/services/maps.service.ts
sorenjohanson: Would that make a big enough difference though? You're still going through all three keys and just moving the callbacks outside of the if, but considering it's only in one method (on the server anyway) I don't see much of a benefit here.

teammapper-backend/src/map/utils/tests/mapFactories.ts
sorenjohanson: Yeah I'm creating a full node now. The other reason I was keen on using Partial was because I'm not using repositories in the factories (yet). Typescript itself doesn't understand that nodeMap is actually nullable, since that's a NestJS property, so it expects both nodeMap and nodeMapId to be present - same with nodeParent. Tests work fine, but I think `mapFactories.ts` could do with a refactor to use repositories.

teammapper-frontend/mmp/src/map/handlers/history.ts
sorenjohanson: Added some now, hopefully clears it up a bit :D

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
JannikStreek: hm ok we can start without it

teammapper-backend/src/map/services/maps.service.ts
JannikStreek: it is not an efficiency/performance refactoring, it wanted to give you feedback on the style of coding and try to avoid unnecessary code duplication (DRY). Currently you have a loop which is coded 3 times with basically the same content. At least for readability, please extract the content where possible into smaller ones.

teammapper-frontend/mmp/src/map/handlers/history.ts
```diff
@@ -26,6 +27,28 @@ export default class History {
         this.snapshots = []
     }
 
+    /**
+     * As snapshots are saved by number (1, 2, 3 etc) and not by node ID, this switches out the number with the affected node ID.
```
JannikStreek, line 31: thank you, this already helps a lot. If you want to go through a list and exchange keys (number, 1,2,3... with the content) with other keys (nodeId), that sounds like a good task for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

I think the code would be easier readable with this. What do you think?

teammapper-backend/src/map/services/maps.service.ts
sorenjohanson: Oh yeah I understand what you were getting at, and I wasn't talking about performance either. For me it doesn't make a massive difference whether the code is in 3 if blocks or 3 call backs, since the (minimal) duplication would still be there. I don't mind changing it if it makes it more readable for you though, since you're the one that's going to have to maintain this in a few months' time :D

teammapper-frontend/mmp/src/map/handlers/history.ts
sorenjohanson: `map()` probably won't work since we're changing the array in place and not creating a new one, but I've tried to simplify the method a bit to hopefully make it a bit more readable.

teammapper-backend/src/map/services/maps.service.ts
sorenjohanson: I've added callbacks now, maybe with a bit of excessive type safety 😂

JannikStreek:
Thanks, coding wise this looks good. I have played around with it a bit.

- Images do not work with "redo". The image appears for the original user again but not for the other users. Therefore, people see different things.
- I noticed that sometimes the positions are different, when a delete is reverted. For instance, when I create three nodes root->A->b->C to the direction of top right and remove those and then undo the deletion. On the original screen the position is the same, but on the other client they show to the left. Basically mirrored to the other side.
- Reverting links doesn't work (or it does, but only on the original client)

sorenjohanson:
The reason images (and probably links) don't work is because the map-sync service currently expects two things: The property that was updated, but by the server-side name (for example, `imageSrc`) and the new property value. Since we're just sending diffs, which are client nodes, it cannot map the property (as `imageSrc` is nowhere to be found).

Will probably have to rework `updateNode`.

sorenjohanson:
Documenting this so I remember it.

I've pushed an update now that basically changes the server-side methods to instead return an MmpNode to the client instead of the usual IMmpClientNode that clients normally work with. Funnily enough, this fixes the issue regarding links and images, but introduces a new issue w.r.t to coordinates, because (and this is the main crux), the `PropertyMapping` we have in [nodes.ts](https://github.com/b310-digital/teammapper/blob/b625df951194c82a8e36f8c5712e7a270eecf430/teammapper-frontend/mmp/src/map/handlers/nodes.ts#L1110) basically works with server-side properties in everything _except_ coordinates, where it suddenly expects a client-side property (on the server side, `coordinatesX` and `coordinatesY` are split, whereas client nodes use `coordinates: { x: 0, y: 0 }`.

I still think we need to rework `updateNode` in this PR to make it work and get rid of server-side properties entirely, but I'll look into this tomorrow.

sorenjohanson:
I've updated `mapChangesUndoRedo` in `map-sync.service.ts` to use a "reverse property mapping" of sorts. This basically mirrors the existing property mapping in `nodes.ts` including oddities such as coordinates and ensures that the existing update process of individual nodes still works across the code. We can still rewrite the `updateNode` method in the future, but we don't _have to_ to make this work.

sorenjohanson:
>I noticed that sometimes the positions are different, when a delete is reverted. For instance, when I create three nodes root->A->b->C to the direction of top right and remove those and then undo the deletion. On the original screen the position is the same, but on the other client they show to the left. Basically mirrored to the other side.

This has been fixed, but I noticed now that there's a (new) bug when pasting nodes, undoing the paste and then redoing the paste. This seems to violate some foreign key constraint, causing the whole app to crash during the `diffAddedCallback`.

sorenjohanson:
There was a race condition in the `diffAddedCallback` thanks to `Promise.all()`, this has been fixed.

teammapper-backend/src/map/utils/tests/mapFactories.ts
```diff
@@ -20,6 +20,35 @@ export const createMmpMap = (overrides = {}): MmpMap => ({
...
+    locked: false,
+    detached: false,
+    k: 0,
+    nodeParentId: "",
+    orderNumber: 0,
+    lastModified: new Date(),
+    createdAt: new Date(),
+    nodeMap: map,
+    nodeMapId: map.id,
+    nodeParent: new MmpNode
```
JannikStreek, line 49: not sure I understand this? can you elaborate what his happening here?

teammapper-frontend/mmp/src/map/handlers/history.ts
```diff
@@ -26,6 +27,26 @@ export default class History {
...
```
JannikStreek, line 31: thanks for the comment. To summarize for me: Now we are converting the numbers to nodeIds as this does not work on the server. But can we change the logic itself to use nodeIds? Do we need to stick with these old numbering system?

teammapper-frontend/mmp/src/map/handlers/history.ts
```diff
@@ -26,6 +27,26 @@ export default class History {
...
+     * @param {MapDiff} snapshotDiff
+     */
+    private switchDiffKeys(snapshotDiff: MapDiff): void {
+        ['added', 'deleted', 'updated'].forEach((key: keyof MapDiff) => {
+            const diffSection = snapshotDiff[key];
+            if (diffSection && typeof diffSection === 'object') {
+                Object.entries(diffSection).forEach(([index, value]: [string, Partial<ExportNodeProperties> | undefined]) => {
+                    const nodeId = value?.id ?? this.snapshots[this.index][Number(index)]?.id;
+                    if (nodeId) {
+                        delete diffSection[Number(index)];
```
JannikStreek, line 42: I would propose a more functional way of doing this. Not modifying an existing data structure in place, but to create a new value without the to deleted value. E.g. by using `filter`.

teammapper-frontend/src/app/core/services/map-sync/map-sync.service.ts
```diff
@@ -429,6 +436,105 @@ export class MapSyncService implements OnDestroy {
       this.mmpService.new(result.map.data, false);
     });
 
+    this.socket.on('mapChangesUndoRedo', (result: ResponseUndoRedoChanges) => {
+      if (result.clientId === this.socket.id) return;
+
+      const ReversePropertyMapping = {
```
JannikStreek, line 442: this should be extracted in a type / constant file. I don't see a reason this has to randomly appear in a function?

JannikStreek:
Cool, couldn't find any bugs for the time being. Let's finalize this and merge it. I have some remaining questions regarding the code, see above.

teammapper-frontend/mmp/src/map/handlers/history.ts
sorenjohanson: The numbering system comes from https://www.npmjs.com/package/deep-object-diff, which makes it hard to change unless you want to dive into the package itself.

teammapper-backend/src/map/utils/tests/mapFactories.ts
sorenjohanson: Regarding `nodeParent`? It's not nullable on either `Node` or `MmpNode` as far as TS is concerned, so there has to be _something_ there (in this case something very ugly :D)

teammapper-backend/src/map/utils/clientServerMapping.ts
```diff
@@ -55,6 +55,31 @@ const mapMmpMapToClient = (
   }
 }
 
+const mergeClientNodeIntoMmpNode = (
```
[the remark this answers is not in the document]
sorenjohanson: I'm not sure the spread operator works here considering that the client node has different keys to the server node (i.e. `colors.name` instead of `colorsName`)? I also wanted to give explicit priority to the client node as that's where the diffs are coming from.

teammapper-backend/src/map/utils/tests/mapFactories.ts
JannikStreek: but its the root, how come its not nullable? We don't have parents for the root note itself?

teammapper-backend/src/map/utils/clientServerMapping.ts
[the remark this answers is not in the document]
JannikStreek: ah ok sorry nvmd

PR #526, opened 2024-10-25
Enable JSON Importer to add missing coordinates

### LINKED ISSUE #524: Enable JSON importer to add missing coordinates
**Proposed feature**
The JSON importer should be able to calculate the nodes of coordinates based on existing helper methods used when adding a new node manually. If the JSON importer is able to do this, we can build other import capabilities on top of it.

sorenjohanson:
closes #524


teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -640,6 +641,77 @@ export default class Nodes {
...
+     * Proactively apply coordinates to all nodes that are missing coordinates
+     * This method is very similar to calculateCoordinates, but it's fully self-contained and only applicable to MapSnapshots.
+     * Adding this functionality to calculateCoordinates would break a lot of things, as Node and ExportNodeProperties are not the same.
+     * @param {MapSnapshot} map
+     * @returns {MapSnapshot} map - Return the map snapshot with properly applied coordinates
+     */
+    public applyCoordinatesToMapSnapshot(map: MapSnapshot): MapSnapshot {
+        const getOrientation = (node: ExportNodeProperties, rootNode: ExportNodeProperties) => node.coordinates?.x < rootNode.coordinates?.x
+        const getLowerNode = (nodes: ExportNodeProperties[]) => {
+            let tmp = nodes[0].coordinates?.y || 0
```
JannikStreek, line 654: This can be improved. Don't use let as a result value and map to find something. There is also `reduce`, `select` etc. to do this. I always try to avoid let and only use const. If you really have to, then map is the wrong iteration function.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -640,6 +641,77 @@ export default class Nodes {
...
+                if (nodeParent && nodeParent.isRoot) {
+                    const rightNodes: Array<ExportNodeProperties> = [],
+                        leftNodes: Array<ExportNodeProperties> = []
+
+                    siblings.map(sibling => {
+                        getOrientation(sibling, rootNode) ? leftNodes.push(sibling) : rightNodes.push(sibling)
+                    })
+
+                    if (leftNodes.length <= rightNodes.length) {
+                        coordinates.x -= 200
```
JannikStreek, line 688: magic numbers should be constants, that you put at the top of the file or somewhere

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: I've refactored the original method and made some adjustments so there's no longer a special version inside `applyCoordinatesToMapSnapshot`, which means the other methods also benefit from the refactoring. Same applies to `getOrientation`.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: These are now defined as constants. I've also applied these constants to `calculateCoordinates`.

sorenjohanson:
I've also added a generic `calculateNodeCoordinates` method to prevent duplication across `calculateCoordinates` and `applyCoordinatesToMapSnapshot`.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -505,10 +509,31 @@ export default class Nodes {
...
+    ): boolean | undefined {
+        // isRoot exists only on Node type, so type guard is needed
+        if ('isRoot' in node && node.isRoot) {
+            return undefined;
+        }
+    
+        // For Node type, use this.getRoot() if rootNode not provided
+        const root = rootNode ?? (('isRoot' in node) ? this.getRoot() : undefined);
+        if (!root) {
+            return undefined;
```
JannikStreek, line 524: just using `return` is already returning undefined?

teammapper-frontend/mmp/src/map/map.ts
```diff
@@ -85,37 +86,100 @@ export default class MmpMap {
      * @return {MmpInstance} mmpInstance
      */
     private createMmpInstance(): MmpInstance {
```
JannikStreek, line 88: we have talked about this, to revert this change and leave it as it was? What did change your opinion that its still this way?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -1084,20 +1159,20 @@ export default class Nodes {
      * @param {boolean} direction
      */
     private moveSelectionOnBranch(direction: boolean) {
-        if ((this.getOrientation(this.selectedNode) === false && direction) ||
-            (this.getOrientation(this.selectedNode) === true && !direction)) {
+        if ((this.getNodeOrientation(this.selectedNode) === false && direction) ||
```
JannikStreek, line 1162: just `!this.getNodeOrientation(this.selectedNode)` or `this.getNodeOrientation(this.selectedNode)` is enough or why the explicit check for true and false? do you want to avoid checking against null or something?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -640,69 +665,119 @@ export default class Nodes {
...
-                coordinates.x += 200
+                coordinates.x += NODE_HORIZONTAL_SPACING;
             }
         }
-
+    
         if (siblings.length > 0) {
-            const lowerNode = this.getLowerNode(siblings)
-            coordinates.y = lowerNode.coordinates.y + 60
+            const lowerNode = 'isRoot' in node
```
JannikStreek, line 713: do we really have to check for this property existing instead of checking if true/false is in the property?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -640,69 +665,119 @@ export default class Nodes {
         }
     }
 
-    /**
-     * Return the appropriate coordinates of the node.
-     * @param {Node} node
-     * @returns {Coordinates} coordinates
-     */
-    private calculateCoordinates(node: Node): Coordinates {
+    private calculateNodeCoordinates(
```
JannikStreek, line 668: I need comments here, I don't understand the params argument here

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,141 @@ export default class Nodes {
...
+     * getSiblings() - Method to get the siblings of the given node
+     * isRoot() - If parent node is root
+     * getOrientation() - Method to get the orientation of the node
+     * @returns 
      */
-    private calculateCoordinates(node: Node): Coordinates {
+    private calculateNodeCoordinates(
+        node: Node | ExportNodeProperties,
+        params: {
+            getParent: () => (Node | ExportNodeProperties) | null,
```
JannikStreek, line 706: do these really need to be functions and not just of type node or something? I think in both cases (only checked getParent and getSiblings) when you are calling this function, you are passing in a lambda that simply calculates a bit later (when it is called):

E.g.
```
getParent: () => node.parent
```

```
const nodeParent = params.getParent();
```

Here, you could simply pass in the parent as a node instead of using a lambda.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,141 @@ export default class Nodes {
...
-                coordinates.x += 200
+                coordinates.x += NODE_HORIZONTAL_SPACING;
             }
         }
-
+    
         if (siblings.length > 0) {
-            const lowerNode = this.getLowerNode(siblings)
-            coordinates.y = lowerNode.coordinates.y + 60
+            const lowerNode = 'isRoot' in node
```
JannikStreek, line 748: and `isRoot === true` ? Is it enough that the property only exists, but being false?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,141 @@ export default class Nodes {
...
             }
         }
-
+    
         if (siblings.length > 0) {
-            const lowerNode = this.getLowerNode(siblings)
-            coordinates.y = lowerNode.coordinates.y + 60
+            const lowerNode = 'isRoot' in node 
+                ? this.getNodeLowerNode(siblings as Node[])
+                : this.getExportLowerNode(siblings as ExportNodeProperties[]);
```
JannikStreek, line 750: these type casts with `as` as far as I know should be avoided and instead types should be inferred. So if it's needed here, the types before maybe are not set correctly? Or why do you need these casts here?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,141 @@ export default class Nodes {
...
+                isRoot: node.parent?.isRoot ?? false,
+                getOrientation: (n: Node) => this.getNodeOrientation(n)
+            }
+        );
+    }
+
+    public applyCoordinatesToMapSnapshot = (map: MapSnapshot): MapSnapshot => {
+        const rootNode = map.find(x => x.isRoot);
+        
+        map.forEach(node => {
```
JannikStreek, line 780: `forEach` is used here on purpose, to work on the existing map and not create a new map e.g. via `map`? If possible, I would prefer creating a new map here - but maybe you need to update all the references.

JannikStreek:
I have tested it, works well with a small example 👍 As discussed, tests can be added later, when the jest pr is merged. These functions would be a good start for the first unit tests.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: I've removed this unnecessarily complex check and just streamlined the types to `(Node | ExportNodeProperties)[]` across the board. Works fine.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: Check has been removed, unnecessarily complex.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: This was previously `map.map`, I don't know why I changed it. This returns a new map now and works properly.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
sorenjohanson: Yeah I think for `getParent` it makes sense, but the other arguments are actually methods (`this.getSiblings()` for example), so it makes sense to keep the lambda there.

sorenjohanson: I've changed `getParent` to `nodeParent` and it's no longer a lambda.

sorenjohanson: Just to clarify regarding `getSiblings`: In the case of a JSON import, it's simply calling `map.filter()`, which is not exactly a method, but in the case of real nodes, it calls `this.getSiblings()` in `nodes.ts`, which requires proper `this` context. `getOrientation` actually calls a method in both cases, hence why I think it makes sense to keep both lambdas there.

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -528,10 +532,31 @@ export default class Nodes {
...
+        const root = rootNode ?? (('isRoot' in node) ? this.getRoot() : undefined);
+        if (!root) {
+            return;
         }
+    
+        return (node.coordinates?.x ?? 0) < (root.coordinates?.x ?? 0);
+    }
+    
+    // Wrapper methods with specific types
+    public getNodeOrientation(node: Node): boolean | undefined {
```
JannikStreek, line 554: why do we need this method?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -528,10 +532,31 @@ export default class Nodes {
...
+        return this.getOrientation(node);
+    }
+    
+    public getExportOrientation(node: ExportNodeProperties, rootNode: ExportNodeProperties): boolean | undefined {
```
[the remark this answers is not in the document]
JannikStreek: same here, you can directly call the other `getOrientation` method?

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,130 @@ export default class Nodes {
...
-                this.getOrientation(sibling) ? leftNodes.push(sibling) : rightNodes.push(sibling)
-            }
-
+            x: nodeParent?.coordinates?.x ?? (node.coordinates?.x ?? 0),
+            y: nodeParent?.coordinates?.y ?? (node.coordinates?.y ?? 0)
+        };
+    
+        let siblings = params.getSiblings();
+    
+        if (nodeParent && params.isRoot) {
```
JannikStreek, line 721: a comment here would be nice what this does

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,130 @@ export default class Nodes {
...
     /**
-     * Return the appropriate coordinates of the node.
-     * @param {Node} node
-     * @returns {Coordinates} coordinates
+     * Base method for calculating node coordinates. 
+     * The reason this exists is so we can work with a JSON snapshot (as given by an import), but also allow saved, "real" nodes to calculate coordinates
+     * This prevents duplication, whilst passing methods that differ depending on whether or not a JSON snapshot or "real" node is calculating coordinates.
+     * @param node Either a node previously saved or one from a JSON snapshot
+     * @param params
+     * getParent() - Method to find the parent of the given node
```
JannikStreek, line 697: needs to be updated, as this is no lambda anymore

teammapper-frontend/mmp/src/map/handlers/nodes.ts
```diff
@@ -664,68 +689,130 @@ export default class Nodes {
...
+            {
+                nodeParent: node.parent,
+                getSiblings: () => this.getSiblings(node),
+                isRoot: node.parent?.isRoot ?? false,
+                getOrientation: (n: Node) => this.getNodeOrientation(n)
+            }
+        );
+    }
+
+    public applyCoordinatesToMapSnapshot = (map: MapSnapshot): MapSnapshot => {
```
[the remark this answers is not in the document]
sorenjohanson: But the `map.map` call was kinda funny :D

PR #565, opened 2024-11-16
Chore: Correct Typescript types in backend entities

### LINKED ISSUE #555: Chore: Correct typescript types in backend entities
**Proposed feature**
Currently the database annotations within typeorm and the given types are not the same. Many fields are nullable but the type checks require not nullable values. These should be the same. See https://github.com/typeorm/typeorm/issues/4139#issuecomment-550307883

sorenjohanson:
closes #555

### DIFF
--- a/teammapper-backend/src/map/entities/mmpMap.entity.ts
+++ b/teammapper-backend/src/map/entities/mmpMap.entity.ts
@@ -11,35 +11,36 @@ import { MmpNode } from './mmpNode.entity'
 @Entity()
 export class MmpMap {
   @PrimaryGeneratedColumn('uuid')
-  id: string
+  id: string;
 
-  @Column({ type: 'timestamptz', nullable: true, default: () => 'now()' })
-  lastModified: Date
+  @Column({ type: 'timestamptz', nullable: true, default: () => 'CURRENT_TIMESTAMP' })
+  lastModified: Date | null = new Date();
 
   @Column({ type: 'timestamptz', nullable: true })
-  lastAccessed: Date
+  lastAccessed: Date | null;
 
-  @Column({ nullable: true })
+  @Column({ type: 'uuid', nullable: true })
   @Generated('uuid')
-  adminId: string
+  adminId: string | null;
 
-  @Column({ nullable: true, default: null })
+  @Column({ type: 'uuid', nullable: true, default: null })
   @Generated('uuid')
-  modificationSecret: string
+  modificationSecret: string | null = null;
 
-  @Column({ nullable: true })
-  name: string
+  @Column({ type: 'varchar', nullable: true })
+  name: string | null;
 
-  @Column('jsonb', { nullable: false, default: {} })
-  options: MapOptions
+  @Column('jsonb', { 
+    nullable: false, 
+    default: {}
+  })
+  options: MapOptions;
 
-  /* eslint-disable @typescript-eslint/no-unused-vars */
-  @OneToMany((type) => MmpNode, (node) => node.nodeMap, {
+  @OneToMany(() => MmpNode, (node) => node.nodeMap, {
     cascade: true,
   })
-  /* eslint-enable @typescript-eslint/no-unused-vars */
-  nodes: MmpNode[]
+  nodes: MmpNode[];
 
-  @Column({ type: 'timestamptz', default: () => 'now()', nullable: true })
-  createdAt: Date
+  @Column({ type: 'timestamptz', nullable: true, default: () => 'CURRENT_TIMESTAMP' })
+  createdAt: Date | null = new Date();
 }
--- a/teammapper-backend/src/map/entities/mmpMap.entity.ts
+++ b/teammapper-backend/src/map/entities/mmpMap.entity.ts
@@ -14,7 +14,7 @@ export class MmpMap {
   id: string;
 
   @Column({ type: 'timestamptz', nullable: true, default: () => 'CURRENT_TIMESTAMP' })
-  lastModified: Date | null = new Date();
+  lastModified: Date | null;
 
   @Column({ type: 'timestamptz', nullable: true })
   lastAccessed: Date | null;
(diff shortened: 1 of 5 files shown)


JannikStreek:
thanks, looking good so far. Can you double check if the db structure stays the same? I see some db types also changed e.g. from float to float8, would be important to see if we need migrations for data and structure.

teammapper-backend/src/map/entities/mmpMap.entity.ts
```diff
@@ -11,35 +11,36 @@ import { MmpNode } from './mmpNode.entity'
 @Entity()
 export class MmpMap {
   @PrimaryGeneratedColumn('uuid')
-  id: string
+  id: string;
 
-  @Column({ type: 'timestamptz', nullable: true, default: () => 'now()' })
-  lastModified: Date
+  @Column({ type: 'timestamptz', nullable: true, default: () => 'CURRENT_TIMESTAMP' })
+  lastModified: Date | null = new Date();
```
JannikStreek, line 17: do we need `new Date()` here? `CURRENT_TIMESTAMP` is already setting the timestamp.

teammapper-backend/src/map/entities/mmpMap.entity.ts
```diff
@@ -11,35 +11,36 @@ import { MmpNode } from './mmpNode.entity'
...
+  @Column({ type: 'uuid', nullable: true })
   @Generated('uuid')
-  adminId: string
+  adminId: string | null;
 
-  @Column({ nullable: true, default: null })
+  @Column({ type: 'uuid', nullable: true, default: null })
   @Generated('uuid')
-  modificationSecret: string
+  modificationSecret: string | null = null;
```
JannikStreek, line 28: same here, why do we need to set it to null if the default is null?

teammapper-backend/src/map/entities/mmpMap.entity.ts
sorenjohanson: As far as I know CURRENT_TIMESTAMP is for Postgres, so `lastModified` would be null by default if you created a new MmpNode() in Typescript.

teammapper-backend/src/map/entities/mmpMap.entity.ts
sorenjohanson: Isn't the default `undefined` if a variable is not set?

teammapper-backend/src/map/entities/mmpMap.entity.ts
JannikStreek: yes, the point is you leave it simply empty until its saved. Thats what the Postgres default is there for. If you assign a default in javascript, you never reach an empty field where the Postgres default comes into play. Why do we need a javascript default?

teammapper-backend/src/map/entities/mmpMap.entity.ts
JannikStreek: depends on how typeorm handles it, you can simply test it yourself. If it's undefined, why not use undefined in the type? You added the type null, we could also use undefined. The parameter in the typeorm annotation `nullable` refers to the database, where `NULL` is the default. So we don't have to use null, we can use undefined. Your pick. But I wouldn't force it to null instead of undefined.

teammapper-backend/src/map/entities/mmpNode.entity.ts
```diff
@@ -17,106 +17,100 @@ import { validateOrReject, IsDefined } from 'class-validator';
...
-  children: MmpNode[]
-  /* eslint-enable @typescript-eslint/no-unused-vars */
+  @OneToMany(() => MmpNode, (node) => node.nodeParent)
+  children: MmpNode[];
 
-  @Column({ default: false })
+  @Column({ type: 'boolean', default: false })
   @IsDefined()
-  root: boolean
+  root: boolean = false;
```
JannikStreek, line 46: thats leftover I guess? :D

teammapper-backend/src/map/entities/mmpNode.entity.ts
```diff
@@ -17,106 +17,100 @@ import { validateOrReject, IsDefined } from 'class-validator';
...
 
-  @Column({ nullable: true })
-  imageSrc: string
+  @Column({ type: 'varchar', nullable: true })
+  imageSrc: string | null;
 
-  @Column({ nullable: true, default: 60 })
-  imageSize: number
+  @Column({ type: 'integer', nullable: true, default: 60 })
+  imageSize: number | null = 60;
```
JannikStreek, line 78: same

teammapper-backend/src/map/entities/mmpNode.entity.ts
```diff
@@ -17,106 +17,100 @@ import { validateOrReject, IsDefined } from 'class-validator';
...
 
-  @Column({ default: false })
+  @Column({ type: 'boolean', default: false })
   @IsDefined()
-  detached: boolean
+  detached: boolean = false;
 
-  @Column({ nullable: true, type: 'float' })
-  k: number
+  @Column({ type: 'float8', nullable: true })
```
JannikStreek, line 90: float8

teammapper-backend/src/map/entities/mmpNode.entity.ts
```diff
@@ -17,106 +17,100 @@ import { validateOrReject, IsDefined } from 'class-validator';
...
 
   @PrimaryColumn('uuid')
   @Index()
   @IsDefined()
-  nodeMapId: string
+  nodeMapId: string;
 
-  @Column({ nullable: true })
-  nodeParentId: string
+  @Column({ type: 'uuid', nullable: true })
```
JannikStreek, line 98: I would propose to use string here instead. https://orkhan.gitbook.io/typeorm/docs/entities#column-types-for-postgres

uuid is a column type in postgres and currently it is a string. This should trigger a database column type change, which is not what we want?

PR #577, opened 2024-11-18
Add additional checks to prevent WsExceptionHandler from throwing undefined errors

### LINKED ISSUE #574: Investigate log errors with WsExceptionsHandler undefined
**Describe the bug**
Investigate causes of this bug, which is present since the undo/redo change (https://github.com/b310-digital/teammapper/commit/5609cdf9355ae3b7f7207c8772b994dfbd42b2e3). Unfortunately, there is not more in the logs. My best guess is an unhandled promise error that is not handled properly. I would suggest to double check the redo/undo pr and double check all promise returns.

Another hint: These are happening multiple times directly after each other, e.g. 7 times this error in the same second.

```
[Nest] 52 - 11/18/2024, 10:15:38 AM ERROR [WsExceptionsHandler] undefined
```

sorenjohanson:
closes #574

Previously, there was no check to see if newNodes actually existed, but since we're returning undefined, `newNodes.length` may actually throw an error.

### DIFF
--- a/teammapper-backend/src/map/controllers/maps.gateway.ts
+++ b/teammapper-backend/src/map/controllers/maps.gateway.ts
@@ -125,7 +125,7 @@ export class MapsGateway implements OnGatewayDisconnect {
       request.mapId,
       request.nodes
     )
-    if (newNodes.length === 0) return false
+    if (!newNodes || newNodes.length === 0) return false
 
     this.server.to(request.mapId).emit('nodesAdded', {
       clientId: client.id,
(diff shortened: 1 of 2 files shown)


sorenjohanson:
Additional fix: The `Promise.reject()`  in `addNodes` was never returned properly, so it's propagated directly upwards to the `WsExceptionHandler` instead of going back to the gateway. It's now returned properly.

JannikStreek:
looking good, thanks!

PR #594, opened 2024-11-27
Solve [WsExceptionsHandler] undefined

### LINKED ISSUE #590: Solve [WsExceptionsHandler] undefined
**Describe the bug**
We still have many `[WsExceptionsHandler] undefined` in the logs as we often return `Promise.reject` without any error. Therefore, the log of the crash is completely useless.

**Expected behavior**
When a node could not be found or similar, we should follow this paradigm: "Exceptions should not be expected". Meaning: If a node can't be found, we should not trigger an exception without a useless error in the log but simply a short log statement and fulfilling the promise with either `undefined`, `false` etc.

For instance:

```
  async updateNode(
    mapId: string,
    clientNode: IMmpClientNode
  ): Promise<MmpNode> {
    const existingNode = await this.nodesRepository.findOne({
      where: { nodeMapId: mapId, id: clientNode.id },
    })

    if (!existingNode) return Promise.reject()
```
Should be changed to simply `return` (meaning a fulfilled promise with `undefined` value). The caller needs to check for undefined then (which already happens in the gateway).

Another example:

```
  async addNodes(
    mapId: string,
    nodes: Partial<MmpNode>[]
  ): Promise<MmpNode[]> {
    if (!mapId || nodes.length === 0) return Promise.reject()
```

Here, returning an empty array would do the trick.

In addition, we should log e.g. that a node can't be found in a specific method.

sorenjohanson:
closes #590


teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -57,10 +59,12 @@ export class MapsGateway implements OnGatewayDisconnect {
   async onJoin(
     @ConnectedSocket() client: Socket,
     @MessageBody() request: IMmpClientJoinRequest
-  ): Promise<IMmpClientMap> {
+  ): Promise<IMmpClientMap | undefined> {
     const map = await this.mapsService.findMap(request.mapId)
-    if (!map)
-      return Promise.reject()
+    if (!map) {
+      this.logger.error(`Could not find map when client ${client.id} tried to join`);
```
JannikStreek, line 65: can you add the map id and the function name in the beginning? the class name is added by the  logger afaik but not the function name.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -47,27 +47,39 @@ export class MapsService {
     this.mapsRepository.update(uuid, { lastAccessed })
   }
 
-  async exportMapToClient(uuid: string): Promise<IMmpClientMap> {
+  async exportMapToClient(uuid: string): Promise<IMmpClientMap | undefined> {
     const map = await this.findMap(uuid)
     if (!map) return Promise.reject(new EntityNotFoundError("MmpMap", uuid))
```
JannikStreek, line 52: I would suggest to standardize our return values. E.g. when a map can't be found always return either rejected promise with error or return a fulfilled one with undefined. As you refactored it to use a fulfilled one, lets also use it here?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -47,27 +47,39 @@ export class MapsService {
...
+      )
+    }
   }
 
-  async addNode(mapId: string, node: MmpNode): Promise<MmpNode> {
+  async addNode(mapId: string, node: MmpNode): Promise<MmpNode | undefined> {
     // detached nodes are not allowed to have a parent
-    if (node.detached && node.nodeParentId) return Promise.reject()
+    if (node.detached && node.nodeParentId) {
+      this.logger.log(`Detached node ${node.id} is not allowed to have a parent.`);
```
JannikStreek, line 71: warn? please also standardize this, when to use warn, when log and when error.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -90,23 +102,28 @@ export class MapsService {
   async addNodesFromClient(
     mapId: string,
     clientNodes: IMmpClientNode[]
-  ): Promise<MmpNode[]> {
+  ): Promise<MmpNode[] | undefined> {
```
JannikStreek, line 105: I would suggest to return an empty array instead of undefined to be consistent?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -90,23 +102,28 @@ export class MapsService {
...
     const mmpNodes = clientNodes.map(x => mapClientNodeToMmpNode(x, mapId))
     return await this.addNodes(mapId, mmpNodes)
   }
 
   async addNodes(
     mapId: string,
     nodes: Partial<MmpNode>[]
-  ): Promise<MmpNode[]> {
-    if (!mapId || nodes.length === 0) return Promise.reject()
+  ): Promise<MmpNode[] | undefined> {
```
JannikStreek, line 113: same here, would suggest empty array?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -90,23 +102,28 @@ export class MapsService {
...
+    }
 
       const reducer = async (previousPromise: Promise<MmpNode[]>, node: MmpNode): Promise<MmpNode[]> => {
         const accCreatedNodes = await previousPromise;
         if (await this.validatesNodeParentForNode(mapId, node)) {
           try {
             const newNode = await this.addNode(mapId, node);
-            return accCreatedNodes.concat([newNode]);
+            if (newNode) {
+              return accCreatedNodes.concat([newNode]);
```
JannikStreek, line 125: be careful to not break the reduce function here - I think a return is missing in the case `newNode` is missing. Maybe just move the return in the catch part one level out of the current scope?

sorenjohanson: I'll return an empty array just in case.

teammapper-backend/src/map/controllers/maps.gateway.ts
```diff
@@ -57,10 +59,12 @@ export class MapsGateway implements OnGatewayDisconnect {
   async onJoin(
     @ConnectedSocket() client: Socket,
     @MessageBody() request: IMmpClientJoinRequest
-  ): Promise<IMmpClientMap> {
+  ): Promise<IMmpClientMap | undefined> {
     const map = await this.mapsService.findMap(request.mapId)
-    if (!map)
-      return Promise.reject()
+    if (!map) {
+      this.logger.error(`onJoin(): Could not find map ${request.mapId} when client ${client.id} tried to join`);
```
JannikStreek, line 65: can you make it warn here like at the other logs? or why here error?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -82,31 +100,38 @@ export class MapsService {
     try {
       return this.nodesRepository.save(newNode)
     } catch(error) {
-      this.logger.error(`${error.constructor.name} - Failed to add node ${newNode.id}: ${error}`)
+      this.logger.error(`${error.constructor.name} addNode(): Failed to add node ${newNode.id}: ${error}`)
```
JannikStreek, line 103: warn?

teammapper-backend/src/map/services/maps.service.ts
JannikStreek: why an empty array? this breaks the whole reduce function due to one potentially broken node. You need to return `accCreatedNodes` here like it is already done in the catch block. You can already reuse the return there if you move it one level down out of the catch block to make it the last line in the function call.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -154,7 +182,7 @@ export class MapsService {
           lastModified: new Date(),
         })
       } catch(error) {
-        this.logger.error(`${error.constructor.name} - Failed to update node ${existingNode.id}: ${error}`)
+        this.logger.error(`${error.constructor.name} updateNode(): Failed to update node ${existingNode.id}: ${error}`)
```
JannikStreek, line 185: warn?

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -226,7 +254,7 @@ export class MapsService {
             try {
               await this.nodesRepository.save(serverNode);
             } catch(error) {
-              this.logger.error(`${error.constructor.name} - Failed to update node ${serverNode.id} during diff update: ${error}`)
+              this.logger.error(`${error.constructor.name} diffUpdatedCallback(): Failed to update node ${serverNode.id}: ${error}`)
```
JannikStreek, line 257: warn?

teammapper-backend/src/map/services/maps.service.ts
sorenjohanson: Yeah good point :D It should now return accCreatedNodes properly.

teammapper-backend/src/map/controllers/maps.controller.ts
```diff
@@ -51,19 +51,23 @@ export default class MapsController {
...
     }
   }
 
   @Post(':id/duplicate')
   async duplicate(
     @Param('id') mapId: string,
-  ): Promise<IMmpClientPrivateMap> {
+  ): Promise<IMmpClientPrivateMap | undefined> {
     const oldMap = await this.mapsService.findMap(mapId).catch((e: Error) => {
       if (e.name === 'MalformedUUIDError') throw new NotFoundException()
```
JannikStreek, line 72: do you want to keep the exception or should we use the same pattern as in the other methods (logging + fulfill with undefined)?

JannikStreek:
Please revert the package.lock ;) otherwise, one last comment and this is ready to be merged 👍

PR #603, opened 2024-12-02
Make updateNode more robust

### LINKED ISSUE #591: Make updateNode more robust
**Proposed feature**
Currently, node updates for non existing nodes are dropped. In theory that sounds good, but in practice if the first request to add the node was dropped, every further request on that node will fail. We should think about creating the node instead if it does not exist during the update. Or even better: Check if an upsert logic would work: https://github.com/typeorm/typeorm/issues/1090.

sorenjohanson:
closes #591

### DIFF
--- a/teammapper-backend/src/map/services/maps.service.ts
+++ b/teammapper-backend/src/map/services/maps.service.ts
@@ -24,7 +24,7 @@ import MalformedUUIDError from './uuid.error'
 @Injectable()
 export class MapsService {
   private readonly logger = new Logger(MapsService.name)
-  
+
   constructor(
     @InjectRepository(MmpNode)
     private nodesRepository: Repository<MmpNode>,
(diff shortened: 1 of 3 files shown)


teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -62,27 +62,24 @@ export class MapsService {
     )
   }
 
-  async addNode(mapId: string, node: MmpNode): Promise<MmpNode> {
+  async addNode(mapId: string, node: MmpNode): Promise<MmpNode | null> {
```
JannikStreek, line 65: Just a quick note: What we always have to check is, if the parent actually exists. Upsert isn't solving this. It's the same for add, update and delete. If the parent was deleted before, it makes no sense to proceed.

JannikStreek: on the other hand, we are already catching errors further below and we can't completely prevent the risk with a check. so as long as we don't have sophisticated syncing mechanisms in place, we maybe can just fallback on that.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -101,21 +98,21 @@ export class MapsService {
...
-          try {
-            const newNode = await this.addNode(mapId, node);
-            return accCreatedNodes.concat([newNode]);
-          } catch (error) {
-            this.logger.warn(`Failed to add node ${node.id} to map ${mapId}: ${error}`);
-            return accCreatedNodes;
-          }
+    const reducer = async (previousPromise: Promise<MmpNode[]>, node: MmpNode): Promise<MmpNode[]> => {
+      const accCreatedNodes = await previousPromise;
+      if (await this.validatesNodeParentForNode(mapId, node)) {
```
JannikStreek, line 103: as we are already catching errors below when the parent is missing, might be ok to remove that here. However, maybe make the catch explicit for certain errors. E.g. if instance of X then catch, otherwise let the error bubble up.

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -140,23 +137,22 @@ export class MapsService {
...
-          ...existingNode,
-          ...mapClientNodeToMmpNode(clientNode, mapId),
-          lastModified: new Date(),
-        })
-      } catch(error) {
-        this.logger.error(`${error.constructor.name} - Failed to update node ${existingNode.id}: ${error}`)
-        return Promise.reject(error)
-      }
+    try {
+      await this.nodesRepository.upsert({
```
JannikStreek, line 144: are you sure upsert does not return the new entity? do we really need the find call?

teammapper-backend/src/map/services/maps.service.ts
sorenjohanson: I don't think we're explicitly checking if the parent node exists, and I'm not sure upsert would throw an error as the parent node can be undefined, or am I missing something?

JannikStreek: I am more referring to a wrong parent Id that does not exist. I think upsert will fail here, maybe do a quick test ?

PR #629, opened 2024-12-16
Add final exception handler to prevent container crashes

### LINKED ISSUE #617: Add final exception handler to prevent container crashes
**Proposed feature**
Currently there's no final exception handler if an exception bubbles up, which causes at the bare minimum the websockets to crash, if not the whole container. We should add one.

sorenjohanson:
closes #617

### DIFF
--- a/teammapper-backend/src/filters/global-exception.filter.ts
+++ b/teammapper-backend/src/filters/global-exception.filter.ts
@@ -1,0 +1,62 @@
+import { ExceptionFilter, Catch, ArgumentsHost, Logger } from '@nestjs/common';
+
+// This is for any gateway and "internal" NestJS related errors - like if the gateway can't reach clients or things like that.
+// It will try to always keep clients and their websockets alive and gracefully send errors over the wire, without revealing internal error reasons.
+@Catch()
+export class GlobalExceptionFilter implements ExceptionFilter {
+  private readonly logger = new Logger(GlobalExceptionFilter.name);
+
+  catch(exception: any, host: ArgumentsHost) {
+    const ctx = host.getType();
+
+    this.logger.error({
+      error: exception,
+      type: exception?.constructor?.name || typeof exception,
+      message: exception?.message || 'Unknown error',
+      stack: exception?.stack,
+      context: ctx,
+    });
+
+    try {
+      switch (ctx) {
+        case 'http': {
+          const response = host.switchToHttp().getResponse();
+          return response.status(500).json({
+            statusCode: 500,
+            message: 'Internal server error',
+            timestamp: new Date().toISOString(),
+          });
+        }
+
+        case 'ws': {
+          const client = host.switchToWs().getClient();
+          const error = {
+            event: 'error',
+            data: {
+              message: 'Internal server error',
+              timestamp: new Date().toISOString(),
+            },
+          };
+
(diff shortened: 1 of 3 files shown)


teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -137,7 +137,8 @@ export class MapsService {
           return accCreatedNodes;
         }
     
-        this.logger.warn(`Parent with id ${node.nodeParentId} does not exist for node ${node.id} and map ${mapId}`);
+        // This needs to be an error because it is almost guaranteed to cause issues down the line due to differing client state
+        this.logger.error(`Parent with id ${node.nodeParentId} does not exist for node ${node.id} and map ${mapId}`);
```
JannikStreek, line 141: lets keep It as a warn, unfortunately this is a rather common problem

teammapper-backend/src/filters/global-exception.filter.ts
```diff
@@ -0,0 +1,62 @@
...
+      error: exception,
+      type: exception?.constructor?.name || typeof exception,
+      message: exception?.message || 'Unknown error',
+      stack: exception?.stack,
+      context: ctx,
+    });
+
+    try {
+      switch (ctx) {
+        case 'http': {
```
JannikStreek, line 22: does it make a difference if it is https or http?

teammapper-backend/src/filters/global-exception.filter.ts
```diff
@@ -0,0 +1,62 @@
+import { ExceptionFilter, Catch, ArgumentsHost, Logger } from '@nestjs/common';
+
+// This is for any unhandled gateway and "internal" NestJS related errors - like if the gateway can't reach clients or things like that.
+// It will try to always keep clients and their websockets alive and gracefully send errors over the wire, without revealing internal error reasons.
+@Catch()
+export class GlobalExceptionFilter implements ExceptionFilter {
+  private readonly logger = new Logger(GlobalExceptionFilter.name);
+
+  catch(exception: any, host: ArgumentsHost) {
```
JannikStreek, line 9: maybe you can find a better type here. If you can't avoid any, use unknown and cast later

teammapper-backend/src/main.ts
```diff
@@ -2,12 +2,29 @@ import { NestFactory } from '@nestjs/core'
...
 async function bootstrap() {
+  const logger = new Logger('Main Process');
+
+  // Process-level handlers for uncaught errors - anything that happens outside of NestJS, such as type errors.
+  // This is only logged server-side so we log the whole stack for better review.
+  process.on('uncaughtException', (error) => {
+    logger.error('Uncaught Exception: ', error.stack);
+  });
+
+  process.on('unhandledRejection', (reason: any) => {
```
JannikStreek, line 17: same, if not known use unknown type - but maybe there is a type available

teammapper-backend/src/map/services/maps.service.ts
```diff
@@ -137,6 +137,7 @@ export class MapsService {
           return accCreatedNodes;
         }
     
+        // This needs to be an error because it is almost guaranteed to cause issues down the line due to differing client state
```
JannikStreek, line 140: the comment is still old? its referring to the log error statement ;)

Issue #677, opened 2025-01-20
You can see when node is dark and text bright, then the link below appears too bright


gnusupport:
![Image](https://github.com/user-attachments/assets/8e19e7ae-715f-4a1d-836a-3c98660640af)

You can see when node is dark and text bright, then the link below appears too bright and not visible as it shares same foreground color.

Additionall, why link is not displayed as link? That is more logical then just displaying the link symbol.

JannikStreek:
> Additionall, why link is not displayed as link? That is more logical than just displaying the link symbol.

The link is shown below to make it still possible to edit the node. Otherwise clicking in the node triggers the link. But maybe you just wanted to update the text.

gnusupport:
I didn't complain on editing of the node. There is main problem in this bug, that is when foreground color is chosen to be white, the symbol of the link below the bubble is going to be invisible, which is accessibility problem.

Maybe there shall be special link color, or program should calculate how not to be invisible. In my particular example it wasn't pure white, that is why you can still somehow see it.

Another accessibility problem is that symbol of the link is rather less meaningful IMHO, as it is not common, so that user can click there is not quite clear. 

It would be good to have at least option that link is shown as link there under the node, instead of the symbol of the link.

If link is www.example.com it would be good having it shown there as such, or the option to have the link be more visible.

![Image](https://github.com/user-attachments/assets/3436056e-1473-42e9-9cd0-b10a9b37375b)

gnusupport:
Not to forget, also the border of the node cannot be customized.

PR #905, opened 2025-07-14
Fix invisible links #677

sguhl:
- Sets the color of links to black
- Adds an option to show linktexts instead of link icons

### DIFF
--- a/teammapper-frontend/mmp/src/map/handlers/draw.ts
+++ b/teammapper-frontend/mmp/src/map/handlers/draw.ts
@@ -293,14 +293,14 @@ export default class Draw {
 
             // Set text or icon based on option
             if (showLinktext) {
-            domText.textContent = this.truncateText(node.link.href);
-            domText.classList.remove('material-icons');
+                domText.textContent = this.truncateText(node.link.href);
+                domText.classList.remove('material-icons');
 
-            domText.style.setProperty('text-decoration', 'underline');
-            domText.style.setProperty('font-style', 'italic');
+                domText.style.setProperty('text-decoration', 'underline');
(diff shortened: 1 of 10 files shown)


teammapper-frontend/src/app/shared/models/settings.model.ts
```diff
@@ -8,6 +8,7 @@ export interface MmpOptions extends OptionParameters {
   fontMaxSize: number;
   fontMinSize: number;
   fontIncrement: number;
+  showLinktext: boolean;
```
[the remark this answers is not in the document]
sguhl: Makes Sense 👍

teammapper-frontend/mmp/src/map/handlers/draw.ts
```diff
@@ -283,17 +283,46 @@ export default class Draw {
...
             const domText = document.createElementNS('http://www.w3.org/2000/svg', 'text')
             domText.textContent = 'link'
             domText.classList.add('link-text')
-            domText.classList.add('material-icons')
-            domText.style.setProperty('fill', DOMPurify.sanitize(node.colors.name))
+
+            // Set text or icon based on option
+            if (showLinktext) {
+                domText.textContent = this.truncateText(node.link.href);
+                domText.classList.remove('material-icons');
```
JannikStreek, line 297: is this remove call needed? in the else clause you are adding this class so I guess its either one too many or it isn't existing here?

sguhl: I agree, but I think with the new refactoring its easier to just always keep it.

PR #953, opened 2025-08-21
Handling of missing assets in nestjs for production

### LINKED ISSUE #922: Handling of missing assets in nestjs for production
**Describe the bug**
Currently, assets are served on production entirely from the angular app, which is configured in nestjs. However, static resources in the asset folder should be not served from angular router but instead directly from nestjs. Also, currently the angular router renders a 404 with status code 200. Maybe a better way handling this is possible.

Example
```
  // Serve static assets first (these take priority)
  app.useStaticAssets(join(__dirname, '..', 'dist/your-angular-app'), {
    prefix: '/assets/',
    setHeaders: (res, path) => {
      // Cache static assets
      if (path.match(/\.(js|css|png|jpg|jpeg|gif|ico|svg|woff|woff2|ttf|eot|map)$/)) {
        res.setHeader('Cache-Control', 'public, max-age=31536000');
      }
    }
  });

  // Serve Angular app files (js, css, etc.)
  app.useStaticAssets(join(__dirname, '..', 'dist/your-angular-app'), {
   ...
  });
```

sguhl:
Serve static assets directly from nestjs instead of using angular router

Closes #922

### DIFF
--- a/teammapper-backend/src/main.ts
+++ b/teammapper-backend/src/main.ts
@@ -50,14 +50,13 @@ async function bootstrap() {
     })
   )
 
-  // Serve static assets first (these take priority)
+  // Serve static assets first
   app.useStaticAssets(join(__dirname, '..', 'client/browser/assets'), {
     prefix: '/assets/',
     setHeaders: (res, path) => {
-      // Cache static assets
       if (
         path.match(
-          /\.(js|css|png|jpg|jpeg|gif|ico|svg|woff|woff2|ttf|eot|map)$/
+          /\.(png|jpg|jpeg|gif|ico|svg|woff|woff2|ttf|eot|map)$/
         )
       ) {
         res.setHeader('Cache-Control', 'public, max-age=31536000')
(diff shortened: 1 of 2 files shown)


teammapper-backend/src/main.ts
```diff
@@ -49,6 +50,26 @@ async function bootstrap() {
...
   )
 
+  // Serve static assets first (these take priority)
+  app.useStaticAssets(join(__dirname, '..', 'client/browser/assets'), {
+    prefix: '/assets/',
+    setHeaders: (res, path) => {
+      // Cache static assets
+      if (
+        path.match(
+          /\.(js|css|png|jpg|jpeg|gif|ico|svg|woff|woff2|ttf|eot|map)$/
```
JannikStreek, line 60: js seems to appear twice, here and in the other rule?

JannikStreek: or maybe I am miss interpreting. So all requests to /assets will be served from here and take priority. All other requests will be redirected to the frontend app.

Still a few things where I would be careful: 
- The cache time seems very high (1 year). What happens if we update the content of a js file? I think the chunks change their file names, not so sure about the static assets. This sounds too high. I would propose a time window of a day. And then force a revalidation.

